PR #7, opened 2015-03-04
Improved readability of bm_app_answer

jin:
Hey all,

I've noticed that most of the benchmark scripts do not have an explanation on what they do. Since the scripts are posted on the rubybench site, it's useful for casual readers to get an idea of what the benchmark is about without having to read the code. 

Made some improvements on one of the scripts here. Let me know what you think. Thanks!

### DIFF
--- a/ruby/benchmarks/bm_app_answer.rb
+++ b/ruby/benchmarks/bm_app_answer.rb
@@ -1,3 +1,8 @@
+# bm_app_answer.rb
+#
+# Use the long-running ackermann function to compute
+# the Answer.
+
 def ack(m, n)
   if m == 0 then
     n + 1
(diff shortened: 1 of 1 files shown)


tgxworld:
I like the idea :+1: 
Since we're going to be generating a digest based on the content of each benchmark file to determine if our benchmark runs are still valid, if we're going to add the descriptions, we'll have to get it right the first time. Otherwise, any new comments or typo fix will result in the invalidation of the existing benchmarks.

@SamSaffron WDYT?

SamSaffron:
I like this, I think we should take ownership of our benchmarking scripts and take pride in them

ruby/benchmarks/bm_app_answer.rb
```diff
@@ -8,8 +13,12 @@ def ack(m, n)
...
+def the_answer_to_life_the_universe_and_everything(m, n)
+  (ack(m, n)
+    .to_s
+    .split(//)
+    .inject(0) { |s, x| s + x.to_i }
+    .to_s + "2").to_i
 end
 
-answer = the_answer_to_life_the_universe_and_everything
+answer = the_answer_to_life_the_universe_and_everything(3, 7) # => 42
```
tgxworld: Not sure if we should change the script here. Will passing variables to methods affect the runtime for this benchmarks?

jin: Good point! I can revert the change since the answer for the method should always return 42.

tgxworld: I'm not sure though. I think for a start we can add the descriptions but lets leave the scripts unchanged first. :smile:

jin: +1. 

This sounds like something that can be benchmarked for.. :stuck_out_tongue:

tgxworld:
Sorry @jin. Going to be closing this as I want the benchmarks to correspond with the benchmarks in the Ruby directory. Thank you for your time :+1:

PR #8, opened 2015-03-11
Rails benchmarks

kirs:
First step of pushing Rails benchmarks.


rails/Gemfile
```diff
@@ -0,0 +1,10 @@
+source 'https://rubygems.org'
+
+gem 'benchmark-ips', github: 'kirs/benchmark-ips', branch: 'quiet-mode'
+gem 'rails', path: '/rails'
```
tgxworld, line 9: Is it possible to set the gem without a path to the local `rails` gem?

Perhaps something like this? `gem 'rails', :git => 'git://github.com/rails/rails.git', :ref => '4aded'`

rails/Gemfile
```diff
@@ -0,0 +1,10 @@
...
+gem 'ffaker'
+gem 'sqlite3'
+gem 'stackprof'
+gem 'mysql2'
+gem 'pg'
+gem 'arel', github: 'rails/arel', branch: 'master'
```
tgxworld: For Gems (ActiveRecord) that may possibly affect the benchmarks, do you think we should lock them to specific versions? Otherwise, we may end up benchmarking a moving target.

rails/benchmarks/activerecord_complex.rb
```diff
@@ -0,0 +1,103 @@
...
+require 'benchmark/ips'
+require 'json'
+
+require 'rails'
+require 'action_controller/railtie'
+require 'active_record'
+require 'sqlite3'
+require 'ffaker'
+
+TIME = (ENV['BENCHMARK_TIME'] || 5).to_i
```
tgxworld: Perhaps we could set all defaults to 10secs? :smile_cat:

rails/benchmarks/activerecord_complex.rb
```diff
@@ -0,0 +1,103 @@
...
+
+def silence_stream(stream)
+  old_stream = stream.dup
+  stream.reopen(IO::NULL)
+  stream.sync = true
+  yield
+ensure
+  stream.reopen(old_stream)
+  old_stream.close
+end
```
tgxworld: I tried using `ActiveRecord::Migration.verbose = false` and it works too :smile:

rails/benchmarks/activerecord_validations.rb
```diff
@@ -0,0 +1,111 @@
...
+require 'action_controller/railtie'
+require 'active_record'
+require 'sqlite3'
+require 'ffaker'
+
+TIME = (ENV['BENCHMARK_TIME'] || 5).to_i
+
+ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
+
+def quietly
```
tgxworld: Ditto https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#r26271915

rails/benchmarks/partials/first.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  Standing on the River Thames, London has been a major settlement for two millennia,
+  its history going back to its founding by the Romans, who named it Londinium.
+  </p>
+
+  <%= render partial: 'second', locals: { posts: posts, users: users } %>
+<% end %>
+
+<%= render collections: posts, partial: 'item' %>
+
+<%= debug @more_records %>
```
tgxworld, line 19: is the `debug` here intentional?

tgxworld:
@kirs I managed to run all the benchmarks so it looks good to me. We can tune the amount of time we run as we go along.

@SamSaffron @schneems Would you like to have a look through as well? :smiley_cat:

rails/Gemfile
kirs, line 9: The ref is set by checkout out to commit here: https://github.com/ruby-bench/ruby-bench-docker/pull/7/files#diff-64e7670fdbf2e48f32d47a260e05e2f3R16
I couldn't find a better solution here.

rails/Gemfile
kirs: We don't need to fix, because Rails master is also built against HEAD gems: https://github.com/rails/rails/blob/master/.travis.yml#L6

rails/benchmarks/partials/first.html.erb
kirs, line 19: Yep

rails/benchmarks/activerecord_complex.rb
kirs: It was 10 firstly, but then I reduced it because waiting wasn't easy. Do you think we will have better results with 10 seconds?

tgxworld: I think 5 or 10 is fine. I just noticed that another script was set to 10, so I figured it'll be good to keep all defaults the same :smile:

rails/Gemfile
tgxworld, line 9: Ah ok! At least for the benchmark-suite, I was wondering if it is possible to remove `path` so that someone who just only has the ruby-bench-suite repo can just run the benchmarks without having to clone `rails`. Not important for now, I can work on it after we merge it.

kirs, line 9: Yeah, when you replace it with `gem 'rails', github: 'rails'` you will be able to run these benchmarks locally.

rails/benchmarks/activerecord_complex.rb
kirs: I've set all the benchmarks to 5 now.

rails/benchmarks/router.rb
```diff
@@ -0,0 +1,177 @@
...
+    "HTTP_USER_AGENT"=>"HTTPie/0.8.0",
+    "rack.version"=>[1, 2],
+    "rack.multithread"=>false,
+    "rack.multiprocess"=>false,
+    "rack.run_once"=>false,
+    "rack.url_scheme"=>"http",
+    "rack.input" => StringIO.new,
+    "HTTP_VERSION"=>"HTTP/1.1",
+    "REQUEST_PATH"=>path
+  }
```
schneems: We could use `Rack::Mock` here instead of manually having to build up the env.

rails/Gemfile
```diff
@@ -0,0 +1,10 @@
+source 'https://rubygems.org'
+
+gem 'benchmark-ips', github: 'kirs/benchmark-ips', branch: 'quiet-mode'
```
schneems: What is different between your `benchmark-ips` and evans? it looks like you're missing some commits he has pushed.

rails/Gemfile
SamSaffron: there are 2 things we are trying to test 
1. Is rails getting faster / slower ... every commit ... in which testing master is appropriate. 
2. Is ruby getting faster / slower ... every commit ... in which we must pin a version

SamSaffron:
the whole way ips is used seems a bit odd to me, we need to easily consume results via a standard api, this is introducing a lot of variability and making consumption of results harder. 

I wonder if it makes sense to break all the benchmarks into standalone files that test one thing and follow the same pattern the ruby benchmarks follow?

rails/Gemfile
kirs: @SamSaffron how you would solve it in the Gemfile?

SamSaffron: At Discourse we use the RAILS_MASTER hack 

https://github.com/discourse/discourse/blob/master/Gemfile

rails/Gemfile
kirs: We need IPS to [support the `quiet` option](https://github.com/evanphx/benchmark-ips/compare/master...kirs:quiet-mode). I will send a PR to the upstream soom.

kirs:
@SamSaffron sure we can, but in some cases it will be slower and increase the code duplication, like in the case with activerecord when we test quite many operations: https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#diff-7ced9d92364c0ac6d54140198f8adda2R99

rails/Gemfile
kirs: Mine has `quiet mode` feature, but I've already sent a PR: https://github.com/evanphx/benchmark-ips/pull/36

SamSaffron: what is the value of quiet mode? how do you access results?

kirs: Because results are collected by parsing JSON [printed by benchmark script](https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#diff-62a3b108fdf213cd836e342593ea887bR157), we want to have JSON-only output.
By default, IPS output is like this:

```
Calculating -------------------------------------
            addition    71.254k i/100ms
           addition2    68.658k i/100ms
           addition3    83.079k i/100ms
addition-test-long-label
                        70.129k i/100ms
-------------------------------------------------
            addition     4.955M (± 8.7%) i/s -     24.155M
           addition2    24.011M (± 9.5%) i/s -    114.246M
           addition3    23.958M (±10.1%) i/s -    115.064M
addition-test-long-label
                         5.014M (± 9.1%) i/s -     24.545M

```

tgxworld:
@kirs Ruby's driver does something like https://github.com/ruby-bench/ruby-bench-suite/blob/master/ruby/benchmarks/driver.rb#L285. Maybe I'll help. :smile_cat:

kirs:
@tgxworld yeah, but with this approach we will have to create 11 files with activerecord benchmarks [instead of one in my solution with IPS](https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#diff-7ced9d92364c0ac6d54140198f8adda2R110).

I guess here are the things I should fix before you merge it:
- [x] use Rack::MockRequest
- [ ] fix Gemfile to be capable with master Rails
- [x] find better way to point on Rails repo instead of `gem 'rails', path: '/rails'`

Is that correct?

SamSaffron:
I think `gem 'rails', path: '/rails'` is actually fine, it makes it very simple to switch the version of rails around. 

That way the docker test runner can simply checkout the tag/commit and then bundle install. 

My overall concern here is that I want us to have a central framework as opposed to many distinct ways of doing benchmarking. 

I think that vision wise the rails benchmarking is far more akin to the standard ruby benches, so there is a lot of appeal in having standalone scripts even if they need a "require" at the top to ensures a central setup is in place. If we can structure stuff like this we can possibly get to a state where we have one central piece for running benches and reuse the framework. 

@tgxworld if the code was merged in as is, how much integration work is needed before we can say, display a graph of "Rails Postgres Model.create" through every stable build of rails since 3.0? 

If we do not go down the "single script for bench"  we are going to lose out big time on the UI eg: http://rubybench.org/ruby/ruby/releases?result_type=app_answer .. no way to display the actual script there cause its just mixed with 20 other things. 

If it was structured like this it would be super simple

`rails_pg_model_create.rb`

```
require 'rails-bench-postgres-setup'

exhibit  = {
    name: ActiveRecord::Faker.name,
    notes: notes,
    created_at: Date.today
}

Benchmark do
   Exhibit.create(exhibit)
end
```

tgxworld:
> @tgxworld if the code was merged in as is, how much integration work is needed before we can say, display a graph of "Rails Postgres Model.create" through every stable build of rails since 3.0?

Hmm even if the script was merged, we'll still need to create the docker image to run the suite https://github.com/ruby-bench/ruby-bench-docker/pull/7. Some challenges I foresee are with regards to the different databases we need to account for. Things like which versions of PostgreSQL, MySQL2.. etc do we freeze it to. Once the image is up, it's pretty trivial to add benchmark for each release and display it on the web front end.

kirs:
> every stable build of rails since 3.0

I didn't realize we're going to run these benchmarks against older commits. What I thought is that it's going to be used only with new commits from master.

@SamSaffron actually your idea of the script is great and I'll try to convert them to this format. Thanks for suggestion :+1:

SamSaffron:
awesome, if you need a bit more of a blank canvas to brainstorm stuff, be
sure to pop on http://community.rubybench.org/ and say hello :)

kirs:
I reworked all the benchmarks :sunglasses: 

1) New format suggested by Sam
2) `DATABASE_URL="sqlite3::memory:" ruby activerecord_scope.rb` allows us to use any DB backend with Activerecord benchmarks without modifying any files

kirs:
@tgxworld @SamSaffron

rails/Gemfile
```diff
@@ -0,0 +1,13 @@
+source 'https://rubygems.org'
+
+gem 'rails', path: '/rails'
+
+# uncomment this to use Rails from HEAD
+# gem 'rails', github: 'rails'
+
+gem 'ffaker'
+gem 'sqlite3'
+gem 'stackprof'
```
tgxworld: I think we can remove `stackprof` for now?

rails/benchmarks/activemodel_views.rb
```diff
@@ -0,0 +1,68 @@
...
+
+class HeavyController < ActionController::Base
+  def index
+    # @more_records = %w(one two three)
+  end
+end
+
+class HeavyView < ActionView::Base
+  # include ApplicationHelper
+  # include Rails.application.routes.url_helpers
```
tgxworld: Do we need the lines that are commented out? Might be better to remove it if we don't need it in order to avoid any confusion.

rails/benchmarks/activerecord_validations_invalid.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'support/activerecord_validations_base.rb'
+
+m = Benchmark.measure do |x|
```
tgxworld: NIT: we can do without `|x|` here.

rails/benchmarks/activerecord_validations_valid.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'support/activerecord_validations_base.rb'
+
+m = Benchmark.measure do |x|
```
tgxworld: Ditto https://github.com/ruby-bench/ruby-bench-suite/pull/8/files#r26854653

kirs:
@tgxworld all fixed!

rails/Gemfile.lock
```diff
@@ -0,0 +1,128 @@
...
+      hike (~> 1.2)
+      multi_json (~> 1.0)
+      rack (~> 1.0)
+      tilt (~> 1.1, != 1.3.0)
+    sprockets-rails (2.2.4)
+      actionpack (>= 3.0)
+      activesupport (>= 3.0)
+      sprockets (>= 2.8, < 4.0)
+    sqlite3 (1.3.10)
+    stackprof (0.2.7)
```
tgxworld: Think we might need to run `bundle install` again xD

rails/benchmarks/app.rb
```diff
@@ -0,0 +1,320 @@
...
+class Comment < ActiveRecord::Base
+  belongs_to :post
+end
+
+# 5.times do
+#   post = Post.create!(title: Faker::Name.name, body: Faker::HipsterIpsum.words(50).join(" "))
+#   2.times do
+#     post.comments.create(email: Faker::Internet.email, body: Faker::HipsterIpsum.words(50).join(" "), author: Faker::Name.name)
+#   end
+# end
```
tgxworld: Opps missed this commented out code :smile_cat:

rails/benchmarks/partials.rb
```diff
@@ -0,0 +1,65 @@
...
+
+class HeavyController < ActionController::Base
+  def index
+    # @more_records = %w(one two three)
+  end
+end
+
+class HeavyView < ActionView::Base
+  # include ApplicationHelper
+  # include Rails.application.routes.url_helpers
```
tgxworld: Same here.

rails/benchmarks/support/activerecord_base.rb
```diff
@@ -0,0 +1,90 @@
+require 'bundler/setup'
+
+require 'active_record'
+require 'json'
+
+RECORDS = (ENV['BENCHMARK_RECORDS'] || 1000).to_i
```
tgxworld: Just a suggestion: Do you think it is better to rename this to `SEED_COUNT`?

tgxworld:
@kirs Awesome work :+1: 

@SamSaffron One concern that I have is with the number of times we're running some code in the benchmarks now that we're not using `benchmark/ips`. I do know that Rails core usually prefer `benchmark/ips` when reporting regressions. I guess we could decide on the number of times to run each benchmark manually too.

kirs:
@tgxworld updated. Thanks!

Concerning number of times to run: running `ruby activerecord_complex.rb` 100 times is not the same as running it 100 times with `benchmarks/ips`. The result will be different because of GC warmups and initializers.

tgxworld:
> Concerning number of times to run: running ruby activerecord_complex.rb 100 times is not the same as running it 100 times with benchmarks/ips. The result will be different because of GC warmups and initializers.

Yup :) What I meant was we have to decide how many `times` we'll have to run the benchmarks now. For instance, we do `10.times` right now for some of the benchmarks. However, is `10.times` enough to produce significant results without too much std deviation? I think `benchmark/ips` solves the above problem but if we decided to not adopt `benchmark/ips`. We'll have to manually decide what `x.times` to use. Let me know if that makes sense.

tgxworld:
@SamSaffron I'll leave it to you to merge this? :smile:

SamSaffron:
I think we need to define a proper helper that we always use for benchmark. 

This helps with a few things ...
1. We can get rid of code duplication hunting out the version of rails
2. It can return json so we don't need to convert stuff to json
3. We can handle disabling of GC (or not) correctly 

EG:

```
require "ruby-bench/benchmark-rails" 

Benchmark.rails(n=1000, "some name perhaps", disable_gc: true, warmup: 3) do
   some_code
end
```

Benchmark by itself is not going to disable gc so results may be very unstable (what happens if GC hits in the middle of measuring?). I am not against using benchmark ips behind the scenes for this stuff but want to make sure we have a clean base interface for this stuff. 

Also, keep in mind that for very short operations the block yielding can end up becoming the majority of cost so you are forced to add while loops (usually not the case for AR testing but worth noting)

tgxworld:
@SamSaffron Will Ruby benchmarks be the same?

SamSaffron:
I merged this in for now, cause the general approach is good it just needs some tweaking, Ruby benches too may be affected by this issue but to a lesser degree, the rails one tend to be micro benchmarks testing a single operation, so GC impact can be huge. 

For the Ruby benches, I guess you kind of want to measure 2 things, 
1. cost of work
2. cost of GC

we can look at improving the way we measure the ruby benches to break that down

kirs:
> the rails one tend to be micro benchmarks testing a single operation, so GC impact can be huge.

Then why did we change it to `Benchmark.measure` with only single run?

tgxworld:
@kirs I think `a single operation` means `Post.all` or `Post.create`. I actually think we should be using `ips` for Rails benchmarks since they benchmarks are meant for the Rails core and they are an advocate of it. In addition, I don't really want to be guessing the number of times we need to run each benchmark. It is alot easier to say that we run all benchmarks for `5` seconds than to find out the `x` number of times we need to run it. Thoughts? :smile:

SamSaffron:
I am fine, but just wrap ips in an interface that better suits us and is
rubybench aware.

PR #13, opened 2015-03-29
Benchmark.rails helper

kirs:
The `Benchmark.rails` helper suggested by @SamSaffron.
It enables warm up before running benchmarks and GC tweaks.

/cc @tgxworld


rails/benchmarks/activemodel_views.rb
```diff
@@ -37,7 +39,7 @@ def user_factory
   { email: Faker::Internet.email }
 end
 
-def render_views
+m = Benchmark.rails(100, "actionview/render_activemodels") do
```
tgxworld, line 42: @kirs Just a thought. Since this script is displayed on the site, Would it be better if we use `kwargs` here? 

`Benchmark.rails(n: 1000, script_name: "actionview/render_activemodels", gc: false)`

tgxworld, line 42: Ah ok I just saw how `Benchmark::rails` is defined. We should use the keys explicitly for readability :)

kirs, line 42: I used kwargs for other arguments and I think it's not necessary for first 2 required args.

rails/benchmarks/app.rb
```diff
@@ -1,290 +1,17 @@
...
-
-m = Benchmark.measure do
-  request(:get, "/posts")
+m = Benchmark.rails(100, "app") do
+  AppBenchmark.request(:get, "/posts")
   2.times do
-    request(:get, "/posts/new")
-    request(:post, "/posts", body: { post: { title: Faker::Food.herb_or_spice, body: Faker::HipsterIpsum.words(50).join(" "), author: Faker::Name.name }})
+    AppBenchmark.request(:get, "/posts/new")
+    AppBenchmark.request(:post, "/posts", body: { post: { title: Faker::Food.herb_or_spice, body: Faker::HipsterIpsum.words(50).join(" "), author: Faker::Name.name }})
```
tgxworld, line 8: Just noticed this. We're going to be measuring the performance of `Faker` here though it might be insignificant.

tgxworld:
@kirs Merging this. 

Do you need anything else for your preparation for your  RailsConf talk?

rails/benchmarks/activerecord_complex.rb
```diff
@@ -39,7 +41,7 @@ class Comment < ActiveRecord::Base
   end
 end
 
-m = Benchmark.measure do
+m = Benchmark.rails(100, "activerecord/#{ENV['DATABASE_URL'].split(":")[0]}/complex") do
```
tgxworld, line 44: @kirs Just wanted to check. What do you pass in as `DATABASE_URL`?

kirs, line 44: I mentioned it here: https://github.com/ruby-bench/ruby-bench-suite/pull/8#issuecomment-84039698

It can accept:
- `sqlite3::memory:`
- `postgres://username:password@localhost/myrailsdb`
- `mysql://user:password@localhost/ruby_bench`

tgxworld, line 44: ah sorry about that missed it :sweat_smile:

kirs:
@tgxworld thanks!

> Do you need anything else for your preparation for your RailsConf talk?

I would like to set Rails benchmarks up and running on rubybench.org. I guess that the next step would be pushing the Docker container.

tgxworld:
@kirs http://rubybench.org/rails/rails/releases?result_type=activerecord/sqlite3_destroy :smile:

kirs:
@tgxworld awesome! What about other adapters than sqlite?

tgxworld:
Currently I only implemented sqlite3. Do you know if the DB will affect the benchmarks significantly? For instance, `save` might be slow on sqlite3 but fast on postgresql

PR #16, opened 2015-04-01
Use Benchmark-ips for Rails benchmarks.

tgxworld:
@SamSaffron @kirs I updated the Rails benchmarks to use `Benchmark-ips`. Thoughts?


SamSaffron:
This is nice less duplicate code! 

We should amend our helper to measure how many objects are allocated for the block, which would be awesome and I would recommend adding. Number of objects allocated is a great measure.

SamSaffron:
To get the count its quite simple..

after ips runs 
1. grab GC.stat[:total_allocated_object]
2. run a single block
3. grab GC.stat[:total_allocated_object]
4. add to json

Keep in mind the name changed here across some ruby versions so account for it.

kirs:
:+1:

rails/benchmarks/support/benchmark_rails.rb
```diff
@@ -11,40 +14,39 @@ def rails(n=1000, label=nil, disable_gc: true, warmup: 3)
...
+    def get_total_allocated_objects
+      if block_given?
+        key =
+          if RUBY_VERSION < '2.2'
+            :total_allocated_object
+          else
+            :total_allocated_objects
+          end
+
+        GC.start
```
tgxworld: Running GC before we count the `total_allocated_objects`/

tgxworld:
@SamSaffron I've included `total_allocated_objects` in the output. :smile:

SamSaffron:
not sure why you are running a GC there, what is the reasoning?

tgxworld:
@SamSaffron Ah ok. While it doesn't matter for `disable_gc: true`, I think we need to run the GC once for `disable_gc: false`. This is to ensure that we don't end up running GC due to the previous allocations from the benchmark. Let me know if that makes sense? :smile_cat:

tgxworld:
@SamSaffron Ping :smiley_cat:

SamSaffron:
sorry, yeah you can merge this, but really no need for gc collect between
the time you start and stop measuring how many object a block allocated.

The counter in GC.stat[:total_allocated_object] is ever increasing, it
never goes done during a processes lifetime

PR #17, opened 2015-04-03
Benchmark JSON API with Rails

kirs:
It's a good idea to benchmark Rails in the JSON API mode too.

PR #18, opened 2015-04-10
Add driver for Rails benchmarks.

tgxworld:


tgxworld:
@SamSaffron @kirs I'm going to try and get these benchmarks (Only for all Rails releases) for Rails up on rubybench.org by this week so @kirs can use it for his talk during RailsConf. After which, I'll have to focus on revision for my exams which is in two weeks :smile:

rails/benchmarks/driver.rb
```diff
@@ -0,0 +1,41 @@
...
+    "benchmark_run[result][total_allocated_objects_per_iteration]" => output["total_allocated_objects_per_iteration"]
+  }
+
+  request.set_form_data({
+    'benchmark_type[category]' => output["label"],
+    'benchmark_type[unit]' => 'iterations per second',
+    'benchmark_type[script_url]' => "#{RAW_URL}#{Pathname.new(path).basename}",
+    'benchmark_run[environment]' => "#{`ruby -v`}",
+    'repo' => 'rails',
+    'organization' => 'ruby'
```
tgxworld, line 35: I'll fix this. Basically rubybench.org works by `rubybench.org/<org name>/<repo name>`. Since the hook is coming from my fork of Ruby, the `<org name>` is originally `tgxworld` but I used a hack that mapped `tgxworld => ruby` to make the URL look nice.

rails/benchmarks/support/app_base.rb
```diff
@@ -5,7 +5,6 @@
 require 'active_record'
 require 'sqlite3'
 require 'ffaker'
-require 'jbuilder'
```
kirs, line 8: This one was required for https://github.com/ruby-bench/ruby-bench-suite/pull/17

kirs:
@tgxworld cool! Let me know if I can help.

tgxworld:
@kirs Could you try running the benchmarks on Rails 3+? I think there are some API changes with how we establish a connection to the database. Passing `sqlite3::memory:` wont work. Thank you!!! :smile:

tgxworld:
Merging this in first. I'm planning to make each benchmark explicit such that people can easily read what is going on directly on RubyBench.org

kirs:
@tgxworld it should work on Rails 3.2+. What version did you try?

tgxworld:
3.0.0 :sweat_smile:

Date: Sat, 11 Apr 2015 00:31:26 -0700

kirs:
I can check it if you really think we need Rails < 3.2, but in my opinion it's not necessary.

PR #19, opened 2015-04-15
Allocation per iteration does not take into account of caches.

tgxworld:
cc/ @SamSaffron @kirs

### DIFF
--- a/rails/benchmarks/support/benchmark_rails.rb
+++ b/rails/benchmarks/support/benchmark_rails.rb
@@ -14,8 +14,10 @@ module Benchmark
         GC.enable
       end
 
-      report = Benchmark.ips(time, warmup, true) do |x|
-        x.report(label) { yield }
+      total_allocated_objects = get_total_allocated_objects do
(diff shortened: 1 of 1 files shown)


kirs:
What kind of cache do you mean?

rails/benchmarks/support/benchmark_rails.rb
```diff
@@ -14,8 +14,10 @@ def rails(label=nil, time:, disable_gc: true, warmup: 3, &block)
         GC.enable
       end
 
-      report = Benchmark.ips(time, warmup, true) do |x|
-        x.report(label) { yield }
+      total_allocated_objects = get_total_allocated_objects do
```
SamSaffron, line 17: not sure if there is value in total allocated objects, per iteration is fine and it happens after warmup anyway.

tgxworld:
Ahh closing this. @SamSaffron's right that we're actually calculating the allocations after warm up and after the benchmarks have been ran.

PR #20, opened 2015-04-16
Add benchmark for ActiveRecord `all` without default scope.

tgxworld:


rails/benchmarks/bm_activerecord_finders_all_with_default_scope.rb
```diff
@@ -0,0 +1,42 @@
...
+    admin: admin
+  }
+
+  User.create!(attributes)
+
+  admin = !admin
+end
+
+Benchmark.rails("activerecord/sqlite3_finders_all_with_default_scope", time: 10) do
+  User.all
```
SamSaffron: I wonder if we should just wack a `to_a` here to make sure 100% we are not returning a relation in any version of rails.

tgxworld: I think thats ok :smile: . I was worried about the performance of `to_a` affecting the benchmarks but if we're using the same Ruby version, it shouldn't affect much.

tgxworld:
Merging this.

PR #22, opened 2015-04-17
Multiple adapters and bundle cache

kirs:
Bundle cache speeds up the code setup from 60+ to 3 seconds :rocket: 
@tgxworld


rails/benchmarks/bm_activerecord_create.rb
```diff
@@ -4,10 +4,7 @@
 
 require_relative 'support/benchmark_rails.rb'
 
-ActiveRecord::Base.establish_connection(
-  adapter: 'sqlite3',
-  database: ':memory:'
-)
+ActiveRecord::Base.establish_connection(ENV['DATABASE_URL'])
```
tgxworld, line 7: Could you try if this works for Rails 3.2+? I was testing it out the other time and was running into problems. Thats why I resorted back to passing `adapter` and `database`.

rails/benchmarks/bm_activerecord_validations_valid.rb
```diff
@@ -1,46 +1,6 @@
...
-  validates :sequence, uniqueness: true
-  validates :age, numericality: { greater_than: 18, less_than: 80 }
-  validates :subdomain, exclusion: { in: %w(www us ca jp),
-    message: "%{value} is reserved." }
-  validates :legacy_code, format: { with: /\A[a-zA-Z]+\z/,
-    message: "only allows letters" }
-  validates :size, inclusion: { in: %w(small medium large),
-    message: "%{value} is not a valid size" }
-end
-
```
tgxworld, line 43: Any reason why this is removed? I moved everything back into a single file without requiring any external file because I wanted the file to be explicit about what it is testing.

rails/benchmarks/driver.rb
```diff
@@ -8,10 +8,15 @@
 
 RAW_URL = 'https://raw.githubusercontent.com/ruby-bench/ruby-bench-suite/master/rails/benchmarks/'
 
+ORMS = [
+  "sqlite3:memory:",
+  "postgres://postgres@#{ENV['POSTGRES_PORT_5432_TCP_ADDR']}:#{ENV['POSTGRES_PORT_5432_TCP_PORT']}/rubybench",
+  "mysql://rubybench@#{ENV['MYSQL_PORT_3306_TCP_ADDR']}:#{ENV['POSTGRES_PORT_5432_TCP_PORT']}/rubybench",
```
tgxworld: :smile_cat: The ENV port is wrong here.

tgxworld:
@kirs Which code setup are you measuring? The benchmarks are still completing in a reasonable amount of time when running on our dedicated server. I'm not really keen on adding/managing a bundler cache on AWS. Thoughts?

rails/benchmarks/bm_activerecord_validations_valid.rb
kirs, line 43: I just wanted to DRY this part between two benchmarks.
Do you want to prefer more duplicated code to make it explicit?

kirs:
> Which code setup are you measuring?

You're right, if the full build time is an hour or so, there is no real speed up from using cache.
However it's still very useful while developing the suite.

rails/benchmarks/bm_activerecord_validations_valid.rb
tgxworld, line 43: Yup I think we should intentionally duplicate code in this case.

screenshot from 2015-04-19 08 53 49

The above makes it easier to read from the site what we're benchmarking. On top of that, the benchmarks are not meant to change often so it'll be fine if we duplicate code here.

tgxworld:
@kirs The build time isn't an hour right now but I think it is in the range of 60+ as you said. No doubt there is going to be a speedup but I don't think maintaining a cache is the responsibility of `ruby-bench-suite`. Perhaps we could do something similar  as https://github.com/ruby-bench/ruby-bench-docker/tree/master/ruby/ruby_releases/discourse_benchmarks/cache? :smile_cat:

rails/benchmarks/bm_activerecord_validations_valid.rb
rafaelfranca, line 43: yeah. seems good to keep it duplicated.

rails/benchmarks/bm_activerecord_create.rb
rafaelfranca, line 7: I think it should work in Rails 3.2, but only if you don't pass any argument at all. I think accepting an URL was implemented only on Rails 4.

rafaelfranca, line 7: Ah. looks it is working. :smile:

tgxworld, line 7: @kirs I tried out the `DATABASE_URL` for sqlite3 and postgresql.

Works fine on postgresql but we need to pass `sqlite3:/:memory` (note the trailing slash) for Rails 3.2+. Other than that this looks good to me. 

@rafaelfranca Thanks for popping by here and reviewing :smile:

kirs, line 7: @tgxworld I already fixed it here https://github.com/kirs/ruby-bench-suite/commit/2e3ca512d7afbfa6a1b35597ab5e3f6b16b74a65

tgxworld, line 7: @kirs Ahh sorry I didn't realized it was a comment on a commit.

kirs:
I'll probably separate these two featured (caching and multi DB benchmarks) into new PRs and improve them.

PR #24, opened 2015-04-19
Fix too many files open issues.

### LINKED ISSUE #2: bm_io_select3 memory benchmark problem
screenshot from 2015-02-06 22 14 09

tgxworld:
Fixes: https://github.com/ruby-bench/ruby-bench-suite/issues/2.

### DIFF
--- a/ruby/benchmarks/bm_io_select2.rb
+++ b/ruby/benchmarks/bm_io_select2.rb
@@ -2,14 +2,10 @@
 
 ios = []
 nr = 1000000
-if defined?(Process::RLIMIT_NOFILE)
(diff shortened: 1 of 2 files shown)


tgxworld:
@SamSaffron Since we need a value that is >= 1024 to be meaningful, I'm fixing max to 4096 which will also be the soft limit on our server. From https://github.com/ruby/ruby/commit/512624b62839f5ffe0faebfceb49411b10ee38bf, it seems like there were similar issues previously. By dividing max by 2, we will ensure that we're never hit the limit. Let me know if this makes sense to you :smile:

PR #28, opened 2015-05-12
Revert "Avoid checking in Gemfile.lock"

tgxworld:
This reverts commit 6dc60d66984e44b84acce5c11a4c3b7d3015360a.

@kirs I made a mistake here. We can't remove Gemfile.lock because some versions of the gems that Rails depends on may differ. Not sure how we should handle this yet. Any ideas?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 rails/vendor/
-Gemfile.lock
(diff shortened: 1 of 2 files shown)


kirs:
What if we create a Gemfile for every Rails version, like `Gemfile.rails32`, `Gemfile.rails42` and so on?
That's what most of the gems do.

tgxworld:
Ah ok. I'll give it a shot to see what I can come up with :)

tgxworld:
Not sure if this is still relevant but we can revisit.

PR #33, opened 2015-06-05
Add new benchmark with to_a

klgilbert:
Added additional benchmark for scope + records.

### DIFF
--- a/rails/benchmarks/bm_activerecord_scope_to_a.rb
+++ b/rails/benchmarks/bm_activerecord_scope_to_a.rb
@@ -1,0 +1,32 @@
+require 'bundler/setup'
+require 'active_record'
+require_relative 'support/benchmark_rails'
+
+ActiveRecord::Base.establish_connection(ENV['DATABASE_URL'])
+
+ActiveRecord::Migration.verbose = false
+
+ActiveRecord::Schema.define do
+  create_table :users, force: true do |t|
+    t.string :name, :email
+    t.timestamps null: false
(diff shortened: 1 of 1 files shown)


klgilbert:
I updated the benchmark with an unique index on email. @SamSaffron is that what you were looking for?

rails/benchmarks/bm_activerecord_scope_to_a.rb
```diff
@@ -0,0 +1,32 @@
...
+
+ActiveRecord::Base.establish_connection(ENV['DATABASE_URL'])
+
+ActiveRecord::Migration.verbose = false
+
+ActiveRecord::Schema.define do
+  create_table :users, force: true do |t|
+    t.string :name, :email
+    t.timestamps null: false
+    t.index(:email, unique: true)
```
tgxworld: Just a nitpick here. Lets leave out the parenthesis for consistency with the lines above.

SamSaffron:
Yeah looks good

Minor change like foobar00% though I am pretty sure the limit clause caps
the cost anyway

klgilbert:
Updated it

SamSaffron:
did you confirm it returns 10 results in an array on local? 

merging it ... lets get it backfilled asap, I want to see the results.

klgilbert:
Yeah it returned an array of 10 results.

tgxworld:
@SamSaffron @klgilbert The benchmark is up.

PR #36, opened 2015-06-28
Make Scaffold app benchmarks work with Rails 3.2+.

tgxworld:


tgxworld:
@kirs Could you review this? :)

cc/ @SamSaffron

rails/benchmarks/views/posts/show.html.erb
```diff
@@ -15,11 +15,5 @@
   <%= @post.author %>
 </p>
 
-<% @post.comments.each do |comment| %>
```
tgxworld, line 18: @kirs I removed the associations first just to keep the benchmarks simple. We could explore further in the future if associations are a key indicator of performance :smile_cat:

tgxworld:
:open_mouth: 

```
{"label":"scaffold_app","version":"3.2.0","iterations_per_second":73.0389464111304,"iterations_per_second_standard_deviation":1.3691325643870598,"total_allocated_objects_per_iteration":16717}

{"label":"scaffold_app","version":"4.2.0","iterations_per_second":32.55812988078875,"iterations_per_second_standard_deviation":3.0714294821646377,"total_allocated_objects_per_iteration":47587}
```

rails/benchmarks/bm_scaffold_app.rb
```diff
@@ -0,0 +1,195 @@
...
+  config.log_level = :debug
+
+  # Send deprecation notices to registered listeners.
+  config.active_support.deprecation = :notify
+
+  # Use default logging formatter so that PID and timestamp are not suppressed.
+  config.log_formatter = ::Logger::Formatter.new
+
+  # otherwise deadlock occurs
+  config.middleware.delete "Rack::Lock"
```
kirs: I disabled logs here with `NullLoger`, because speed of log writing may depend on SSD/HDD of the computer.
What do you think about disabling logs for benchmark?

kirs:
Looks good for me!

rails/benchmarks/bm_scaffold_app.rb
tgxworld: Ah icic. Makes sense to me :) I'll revert my changes.

rails/benchmarks/bm_scaffold_app.rb
```diff
@@ -0,0 +1,195 @@
...
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_post
+    @post = Post.find(params[:id])
+  end
+
+  # Never trust parameters from the scary internet, only allow the white list through.
+  def post_params
+    params.require(:post).permit(:title, :body, :author)
```
tgxworld: Hmm just found out `strong_parameters` is not included in Rails 3.2+.

I'm starting to wonder if there is merit in benchmarking Rails 3.2 for all the work that we have to do in order to take into account of the differences. @SamSaffron WDYT?

PR #40, opened 2015-07-22
Add digest to benchmark drivers

tgxworld:


bundler/benchmarks/driver.rb
```diff
@@ -70,6 +72,10 @@ def run_single(path)
     puts "Posting results to Web UI...."
   end
 
+  def generate_digest(path)
+    Digest::SHA2.hexdigest("#{File.read(path)}#{`ruby -v`}")
```
tgxworld: @segiddins @indirect Do let me know if there are more dependencies that we should include in the digest. Many thanks! :smile:

rails/benchmarks/bm_scaffold_create.rb
```diff
@@ -4,13 +4,6 @@
 require 'active_record'
 require_relative 'support/benchmark_rails'
 
-ENV['RAILS_ENV'] = 'production'
-
-class NullLogger < Logger
-  def initialize(*args);end
-  def add(*args, &block);end
-end
```
tgxworld, line 12: @SamSaffron @kirs We have to log because logging do affect performance quite a bit. Setting to a NullLogger was not running through any of the logging code.

rails/benchmarks/driver.rb
```diff
@@ -7,26 +7,19 @@
...
-    'sqlite3::memory:'
-  end
-
 postgres_tcp_addr = ENV['POSTGRES_PORT_5432_TCP_ADDR'] || 'localhost'
 postgres_port = ENV['POSTGRES_PORT_5432_TCP_PORT'] || 5432
 mysql_tcp_addr = ENV['MYSQL_PORT_3306_TCP_ADDR'] || 'localhost'
 mysql_port = ENV['MYSQL_PORT_3306_TCP_PORT'] || 3306
 
-DATABASE_URLS = [
-  sqlite3_url,
```
tgxworld, line 26: @kirs @SamSaffron I dropped support for sqlite3 here. I figured that the results are not really useful for majority of the Rails apps out there. Less benchmarks to run == backfill faster :stuck_out_tongue:

tgxworld:
Merging in a few days if there aren't any major objections :smile_cat:

bundler/benchmarks/driver.rb
segiddins: Maybe rubygems version as well?

tgxworld: @segiddins ah yes! Will add it in :smile:

rails/benchmarks/bm_scaffold_create.rb
kirs, line 12: I've used NullLogger here to avoid artefacts after benchmark run. If we cleanup, there is no problem to use a classic Rails log

rails/benchmarks/driver.rb
kirs, line 26: :+1:

rails/benchmarks/bm_scaffold_create.rb
tgxworld, line 12: @kirs I think it should be fine since we run the benchmarks in containers that we throw away after each run. :smile_cat:

discourse/benchmarks/bench.rb
```diff
@@ -173,6 +173,10 @@ def bench(path)
   percentiles
 end
 
+def generate_digest
+  Digest::SHA2.hexdigest("#{ENV['POSTGRES_ENV_PG_VERSION']}#{ENV['REDIS_ENV_REDIS_VERSION']}")
```
tgxworld: TODO: Add Discourse commit hash.

http://community.rubybench.org/t/discourse-keeps-changing-can-we-freeze-the-version-of-discourse-bench/42

PR #44, opened 2015-08-19
JRuby releases

klgilbert:
Should I be copying the benchmarks into the JRuby folder or should I just symlink/copy them before running the benchmark? The only files that needed to be changed were driver.rb and memory_driver.rb.


tgxworld:
Symlinking or changing the path in the driver should be the way to go. Duplicating all the benchmarks is going to add overhead in maintainence :smile:

klgilbert:
Yeah I thought it would be wasted effort, but I wasn't sure what you wanted.

klgilbert:
That should do it

Issue #57, opened 2016-02-16
Interest in benchmark against ActionController::API?


bf4:
I would write one to return JSON.  

I could also write one with ActiveModelSerializers or JSONAPI::Resources, both in the conversation about Rails 5.1

tgxworld:
Yup that would be great :) Let me know if you need any help!

PR #60, opened 2016-05-06
Add ActionController::API benchmark with Kir

bf4:
Resolves https://github.com/ruby-bench/ruby-bench-suite/issues/57

(This is a first draft, intended to be improved)

### DIFF
--- a/rails/benchmarks/bm_actioncontroller_api.rb
+++ b/rails/benchmarks/bm_actioncontroller_api.rb
@@ -1,0 +1,16 @@
+require 'bundler/setup'
+require 'action_controller'
+require_relative 'support/benchmark_rails'
+
+class ApiController < ActionController::API
+  def show
+    render json: "JSON", status: :ok
+  end
+end
+app = ApiController.action(:show)
+request = Rack::MockRequest.new(app)
+Benchmark.rails("actioncontroller/api", time: 5) do
+  response = request.get("/")
+  fail "Bad response #{response.status}" if response.status != 200
+  fail "Bad response #{response.body}" if response.body != "JSON"
+end


kirs:
@tgxworld check it out. We paired on this one on Railsconf

rails/benchmarks/bm_actioncontroller_api.rb
```diff
@@ -0,0 +1,16 @@
...
+  def show
+    render json: "JSON", status: :ok
+  end
+end
+app = ApiController.action(:show)
+request = Rack::MockRequest.new(app)
+Benchmark.rails("actioncontroller/api", time: 5) do
+  response = request.get("/")
+  fail "Bad response #{response.status}" if response.status != 200
+  fail "Bad response #{response.body}" if response.body != "JSON"
```
bf4, line 15: Obviously there's more tests that we could add, this should be a good starting point, no?

bf4, line 15: Maybe add a second controller test here that using AC::Base?

tgxworld:
Hi @bf4, sorry I missed this.

I'm just wondering if there is anything different that `ActionController::API` does?

bf4:
@tgxworld yes, it's a smaller middleware stack and has slightly different error handling.  The smaller stack is mostly for performance reasons, removing parts not needed for an API.

rails/benchmarks/bm_actioncontroller_api.rb
kirs, line 15: I don't think we need more tests these two.

tgxworld:
Thanks @bf4 @kirs!

rails/benchmarks/bm_actioncontroller_api.rb
```diff
@@ -0,0 +1,16 @@
+require 'bundler/setup'
+require 'action_controller'
+require_relative 'support/benchmark_rails'
+
+class ApiController < ActionController::API
+  def show
+    render json: "JSON", status: :ok
```
bf4, line 7: curiously enough, "JSON" isn't actually valid JSON (unless you're using some quirky mode processing like rails does by default). Maybe we should change this to `render plain: "TEXT", status: :ok`?

PR #62, opened 2016-05-08
Simple sprockets benchmark

kirs:
cc @tgxworld


rails/benchmarks/assets/javascripts/one.js
```diff
@@ -0,0 +1 @@
+alert()
```
tgxworld: Perhaps using JQuery here might represent a real world use case better? :)

PR #64, opened 2016-05-18
Fix bundler lib env var

segiddins:


tgxworld:
Thanks @segiddins ! I'll rerun the benchmarks :)

PR #65, opened 2016-06-12
Sequel benchmark to create string columns

shahsaurabh0605:


rails/benchmarks/bm_sequel_create_string_columns.rb
```diff
@@ -0,0 +1,31 @@
...
+
+COUNT=25 
+
+Sequel.migration do
+  change do
+    create_table(:users) do
+      primary_key :id
+      COUNT.times do |i|
+        column :"column#{i}", "varchar(255)"
+      end
```
tgxworld: I think we are missing the time stamp columns here.

rails/benchmarks/bm_sequel_create_string_columns.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+end
+
```
tgxworld, line 17: Small nitpick, extra line here

tgxworld:
What output do you get when you run the benchmark file?

shahsaurabh0605:
Had a tough time solving the errors to get the output, so finally:
`{"label":"sequel/sqlite_create_string_columns","version":"5.1.0.alpha","iterations_per_second":1638.2596193728432,"iterations_per_second_standard_deviation":6.653402104956189,"total_allocated_objects_per_iteration":654}
`

tgxworld:
@shahsaurabh0605 What errors were you getting? :thought_balloon: You need to make sure you can run each benchmark file.

shahsaurabh0605:
To run the sequel benchmarks I have to run `RAILS_MASTER=1 DATABASE_URL=sqlite::memory: ruby benchmarks/bm_sequel_create_string_columns.rb`. So the adapter in `DATABASE_URL` is just `sqlite` instead of `sqlite3`. This lead to further complications but now its running.

tgxworld:
Did you try with `postgres` and `mysql`? You can have a look at 

https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/benchmarks/driver.rb#L14-L17
to see how it is done.

shahsaurabh0605:
I looked into that but could not figure out how to use it. What should I write in DATABASE_URL?

shahsaurabh0605:
In the case of mysql I need to set the default `:null=>true` for time stamp columns. I get this output:
`{"label":"sequel/mysql2_create_string_columns","version":"5.1.0.alpha","iterations_per_second":21.765204440161668,"iterations_per_second_standard_deviation":18.377957399835427,"total_allocated_objects_per_iteration":320}
`

tgxworld:
> In the case of mysql I need to set the default :null=>true for time stamp columns.

Why is this so :question: 

What about PostgreSQL?

shahsaurabh0605:
For memory testing in mysql I had to create table `users` to run the benchmark file which initially has null in `created_at` and `updated_at` fields. So for testing I kept null=true. This problem won't happen in postgres. In postgres I get this output:
`{"label":"sequel/postgres_create_string_columns","version":"5.1.0.alpha","iterations_per_second":89.41151072618388,"iterations_per_second_standard_deviation":4.4736969183416475,"total_allocated_objects_per_iteration":353}`

tgxworld:
Awesome! Merging in :)

PR #66, opened 2016-06-14
Add throw/catch benchmark

bf4:
A casual benchmark shows me throw/catch is 2x faster that raise/rescue.
If figure this should be tracked.

https://gist.github.com/bf4/5a576b9a46b4c63a8fe942dbc47aaba8

``` ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.report(:raise_rescue) do
    begin
      raise ArgumentError
    rescue ArgumentError
    end
  end

  x.report(:throw_catch) do
    begin
      catch(:halt) do
        throw(:halt)
      end
    end
  end
  x.compare!
end

  # Ruby 2.2.4 on OSX
  # Warming up --------------------------------------
  #         raise_rescue    58.744k i/100ms
  #          throw_catch   157.030k i/100ms
  # Calculating -------------------------------------
  #         raise_rescue    704.836k (± 6.7%) i/s -      3.525M in   5.023761s
  #          throw_catch      2.532M (± 5.3%) i/s -     12.719M in   5.037449s
  #
  # Comparison:
  #          throw_catch:  2532310.2 i/s
  #         raise_rescue:   704835.7 i/s - 3.59x slower
```

### DIFF
--- a/ruby/benchmarks/bm_app_throw.rb
+++ b/ruby/benchmarks/bm_app_throw.rb
@@ -1,0 +1,7 @@
+i = 0
+while i<300000
+  i += 1
+  catch(:ball) do
+    throw(:ball)
+  end
+end


tgxworld:
Hi @bf4! Thank you for the PR. However, I think this benchmark is more suitable for https://github.com/JuanitoFatas/fast-ruby. 

When we add benchmarks, we are not trying to find out if writing code in different ways will be faster. Instead, we want benchmarks that cover the common use cases and monitor it to see if it becomes slower.

bf4:
@tgxworld Thanks.  Might I ask how it is different from https://github.com/ruby-bench/ruby-bench-suite/blob/5173fc22d3831a22ea815d8357d1e08bfeacc673/ruby/benchmarks/bm_app_raise.rb ? 

In other news, @SamSaffron had suggested that

> diff would be much larger if you had a big backtrace (I think) and Rails starts with 150 deep backtraces [tweet](https://twitter.com/samsaffron/status/742564943333974018)

though when I wrote

> Think it's worth adding that to the benchmark? (Discussing using exceptions app vs auth middleware for 401s)

he responded

>  401s should be very rare, I don't think there is too much need to optimize there, middleware crawl prob 15x slower than exception [tweet](https://twitter.com/samsaffron/status/742837877625622529)

So, perhaps there should be a benchmark showing that 
- throw/catch in a middleware crawl is 15x slower than raise/rescue to the exceptions app
- throw/catch can be > 2x faster than raise/rescue when there are hundreds of backtraces

tgxworld:
OO sorry @bf4. I totally missed the file changes. I thought we were adding the benchmark script in the PR's comment. The file you added is good :+1:

bf4:
Thanks. What do you think about the other scenarios I brought up in https://github.com/ruby-bench/ruby-bench-suite/pull/66#issuecomment-226652132 ?

PR #67, opened 2016-06-15
Sequel benchmarks for rails

shahsaurabh0605:
I have not included the sequel equivalent of `bm_activerecord_scope_all_with_default_scope.rb` because there is no such concept of `default_scope` in sequel.
Also, sequel doesn't provide the `exclusion` helper, so I removed it from the validations.

### DIFF
--- a/rails/benchmarks/bm_sequel_finders_find_by_attributes.rb
+++ b/rails/benchmarks/bm_sequel_finders_find_by_attributes.rb
@@ -1,0 +1,30 @@
+require 'bundler/setup'
+require 'sequel'
+require_relative 'support/benchmark_rails'
+
+DB = Sequel.connect(ENV.fetch('DATABASE_URL'))
+
+DB.create_table(:users) do
+  primary_key :id
+  column :name, "varchar(255)"
+  column :email, "varchar(255)"
+  column :created_at, "timestamp"
+  column :updated_at, "timestamp"
+end
+
+class User < Sequel::Model; end
+
+attributes = {
+  name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+  email: "foobar@email.com"
+}
+
+1000.times do
+  User.create(attributes)
+end
+
+User.create(name: 'kir', email: 'shatrov@me.com')
+
+Benchmark.rails("sequel/#{db_adapter}_finders_find_by_attributes", time: 5) do
+  User.find(email: 'shatrov@me.com')
+end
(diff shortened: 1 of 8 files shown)


rails/benchmarks/bm_sequel_finders_find_by_attributes.rb
```diff
@@ -0,0 +1,30 @@
+require 'bundler/setup'
+require 'sequel'
+require_relative 'support/benchmark_rails'
+
+DB = Sequel.connect(ENV.fetch('DATABASE_URL'))
+
+DB.create_table(:users) do
+    primary_key :id
```
tgxworld: Nitpick: Two spaces for indentation :)

rails/benchmarks/bm_sequel_finders_find_by_attributes.rb
```diff
@@ -0,0 +1,30 @@
...
+}
+
+1000.times do
+  User.create(attributes)
+end
+
+User.create(name: 'kir', email: 'shatrov@me.com')
+
+Benchmark.rails("sequel/#{db_adapter}_finders_find_by_attributes", time: 5) do
+  User.find(:email=>'shatrov@me.com')
```
tgxworld: Nitpick: prefer `email: 'shatrov@me.com'`

shahsaurabh0605:
Done :+1:

tgxworld:
@shahsaurabh0605 Good job :+1: . Before I review. Can you make sure we can run all sequel benchmarks using `driver.rb`?

shahsaurabh0605:
All the sequel benchmarks are running provided I keep `null: true` for timestamp columns. Should I make that change?

tgxworld:
@shahsaurabh0605 Ok I missed this. We have to move sequel away from the `rails` folder. Anything in the rails folder means it depends on Rails. Can you move the sequel benchmarks into it's own folder like bundler? We'll benchmark sequel across releases and on a per commit basis.

shahsaurabh0605:
I am not sure on the dependencies as in what the other directory will contain. Anyways I will try and submit a PR.

tgxworld:
Each folder correspond to a particular repository. So having sequel in the Rails folder is a little confusing because we're benchmarking Sequel against a moving Rails repository. What version should Sequel be fixed to? Instead, we'll just benchmark Sequel and have the ability to compare different benchmarks across repository on RubyBench.org

PR #68, opened 2016-06-16
Add Sequel benchmarks

shahsaurabh0605:
Added the sequel benchmarks away from the rails  folder. Review it for the changes to be done.


sequel/Gemfile
```diff
@@ -0,0 +1,44 @@
+source 'https://rubygems.org'
+
+module ::Kernel
+  def rails_master?
+    ENV["RAILS_MASTER"] == '1'
+  end
+
+  def mysql2_prepared_statements?
+    ENV["MYSQL2_PREPARED_STATEMENTS"] == '1'
```
tgxworld, line 9: Are we able to enable prepared statement in sequel? This flag is used for that.

sequel/Gemfile
```diff
@@ -0,0 +1,44 @@
+source 'https://rubygems.org'
+
+module ::Kernel
+  def rails_master?
+    ENV["RAILS_MASTER"] == '1'
```
tgxworld: We don't need Rails master here. We should fix Rails to the latest stable version. Sequel should be the only variable here.

sequel/Gemfile
```diff
@@ -0,0 +1,44 @@
...
+ # gem 'rack', github: 'rack/rack', branch: 'master'
+ # gem 'sprockets-rails', github: 'rails/sprockets-rails', branch: 'master'
+ # gem 'sprockets', github: 'rails/sprockets', branch: 'master'
+ # gem 'jquery-rails', github: 'rails/jquery-rails', branch: 'master'
+ # gem 'mail', github: 'mikel/mail', branch: 'master'
+ # gem 'globalid', github: 'rails/globalid', branch: 'master'
+ # gem 'sass-rails', github: 'rails/sass-rails', branch: 'master'
+ # gem 'turbolinks', github: 'turbolinks/turbolinks-rails', branch: 'master'
+else
+  gem 'rails', path: '/rails'
```
tgxworld: The sequel gem is missing here if `rails_master?` is `false`

sequel/Gemfile
```diff
@@ -0,0 +1,44 @@
...
+end
+
+gem 'sqlite3', '1.3.10'
```
tgxworld: Is there any where we are using sqlite3 for the sequel benchmarks?

sequel/Gemfile
```diff
@@ -0,0 +1,44 @@
...
+else
+  gem 'mysql2', '0.3.18'
+end
+
+gem 'pg', '0.18.1'
+gem 'benchmark-ips', '~> 2.2.0'
+gem 'em-hiredis', '~> 0.3.0'
+gem 'redis', '~> 3.0'
+gem 'faye-websocket', '~> 0.10.2'
+gem 'puma', '~> 2.16.0'
```
tgxworld: Which of the above gems do we actually need?

sequel/benchmarks/bm_sequel_validations_valid.rb
```diff
@@ -0,0 +1,52 @@
...
+end
+
+class Post < Sequel::Model
+  plugin :validation_helpers
+  def validate
+    super
+    validates_presence :title
+    validates_unique :sequence
+    validates_numeric :age, numericality: { greater_than: 18, less_than: 80 }
+    validates_format /\A[a-zA-Z]+\z/, :legacy_code, :message=> "only allows letters"
```
tgxworld: Prefer `message: "only allows letters"` over `:message => "only allows letters"`

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,194 @@
...
+  def initialize(options)
+    @repeat_count = options[:repeat_count]
+    @pattern = options[:pattern]
+  end
+
+  def run
+    files.each do |path|
+      next if !@pattern.empty? && /#{@pattern.join('|')}/ !~ File.basename(path)
+
+      if path.match(/sequel/)
```
tgxworld: This driver is only responsible for running sequel benchmarks. Do we need this conditional statement?

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,194 @@
...
+    data = {
+      'benchmark_type[category]' => output["label"],
+      'benchmark_type[script_url]' => "#{RAW_URL}#{Pathname.new(path).basename}",
+      'benchmark_type[digest]' => generate_digest(path, database),
+      'benchmark_run[environment]' => "#{`ruby -v`}",
+      'repo' => 'sequel',
+      'organization' => 'sequel'
+    }
+    
+    if(ENV['RAILS_COMMIT_HASH'])
```
tgxworld: We are benchmarking sequel here. So Rails shouldn't be the variable. Instead this should be `SEQUEL_COMMIT_HASH`

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,194 @@
...
+      data['commit_hash'] = ENV['RAILS_COMMIT_HASH']
+    elsif(ENV['RAILS_VERSION'])
```
tgxworld: Ditto https://github.com/ruby-bench/ruby-bench-suite/pull/68/files#r67524584

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  def endpoint
+    @endpoint ||= begin
+      http = Net::HTTP.new(ENV["API_URL"] || 'rubybench.org', 443)
+      http.use_ssl = true
+      http
+    end
+  end
+
+  def generate_digest(path, database)
```
tgxworld, line 139: This method is use to generate a digest based on the versions of the dependencies we are using. This is to ensure our benchmarks are based off the same versions and if they are not, we will invalidate the benchmark. What you need to do here is to think about the dependencies and add them to the string for generating the digest.

sequel/benchmarks/support/benchmark_rails.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      report = Benchmark.ips(time, warmup, true) do |x|
+        x.report(label) { yield }
+      end
+
+      entry = report.entries.first
+
+      output = {
+        label: label,
+        version: ::Rails.version.to_s,
```
tgxworld: This should probably be the version of Sequel currently being benchmarked.

sequel/benchmarks/driver.rb
shahsaurabh0605, line 139: I think we are benchmarking sequel for a specific version of rails, so rails comes off as a dependency. I cannot think of any other dependency for sequel.

sequel/Gemfile
shahsaurabh0605, line 9: Yes, we are able to enable prepared statements for sequel.

sequel/Gemfile
shahsaurabh0605: For running benchmarks locally, we need `sqlite3` gem for DATABASE_URL.

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,188 @@
...
+    string = "#{File.read(path)}#{`ruby -v`}#{`rails -v`}"
+    puts string
```
tgxworld: stray `puts`.

sequel/Gemfile
```diff
@@ -0,0 +1,29 @@
...
+end
+
+if sequel_master?
+  gem 'sequel', github: 'jeremyevans/sequel', branch: 'master'
+else
+  gem 'sequel', path: '/sequel'
+end
+
+gem 'rails', '~> 5.0.0.rc1'
+gem 'sqlite3', '1.3.10'
```
tgxworld: Do we need sqlite3?

sequel/Gemfile
```diff
@@ -0,0 +1,29 @@
...
+  end
+end
+
+if sequel_master?
+  gem 'sequel', github: 'jeremyevans/sequel', branch: 'master'
+else
+  gem 'sequel', path: '/sequel'
+end
+
+gem 'rails', '~> 5.0.0.rc1'
```
tgxworld: I think we only need `ActiveRecord` here.

sequel/benchmarks/bm_sequel_destroy.rb
```diff
@@ -0,0 +1,26 @@
...
+class User < Sequel::Model; end
+
+attributes = {
+  name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+  email: "foobar@email.com"
+}
+
+Benchmark.sequel("sequel/#{db_adapter}_destroy", time: 5) do
+  # we need to create the record in order to delete it
+  user = User.create(attributes)
```
tgxworld, line 27: For sequel, we need [raise_on_save_failure](http://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/ClassMethods.html). Otherwise, saving a record may fail giving us an inaccurate benchmark

sequel/benchmarks/bm_sequel_validations_invalid.rb
```diff
@@ -0,0 +1,43 @@
...
+
+DB.create_table(:posts) do
+  primary_key :id
+  column :title, "varchar(255)"
+  column :author, "varchar(255)"
+  column :body, "Text"
+  column :sequence, 'integer'
+  column :age, 'integer'
+  column :legacy_code, "varchar(255)"
+  column :size, "varchar(255)"
```
tgxworld: > Columns are generally created by specifying the column type as the method name

Can you update all the columns to use the column type as the name instead? Thanks! :)

sequel/benchmarks/support/benchmark_rails.rb
```diff
@@ -0,0 +1,55 @@
+require 'rails'
```
tgxworld: I don't think this is needed?

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,188 @@
...
+    request.set_form_data(form_data.merge(results))
+    endpoint.request(request)
+  end
+
+  def files
+    Dir["#{File.expand_path(File.dirname(__FILE__))}/*"].select! { |path| path =~ /bm_.+/ }
+  end
+
+  def run_single(path, connection: nil, database: nil)
+    script = "SEQUEL_ENV=production ruby #{path}"
```
tgxworld: `SEQUEL_ENV` doesn't seem to be used in sequel anywhere unless I am missing something?

sequel/Gemfile
shahsaurabh0605: Adding only activerecord gives the error:
`block in replace_gem': railties is not part of the bundle. Add it to Gemfile. (Gem::LoadError)` 
So I added `railties` to the gemfile as well and then its working!

sequel/Gemfile
shahsaurabh0605: For running benchmarks locally, we need `sqlite3` gem for DATABASE_URL

sequel/Gemfile
tgxworld: Ok then do that :) We don't need the entire Rails library

sequel/Gemfile
tgxworld: If we're running it locally, we should expect the user to have MySQL and PG set up.

sequel/benchmarks/bm_sequel_destroy.rb
shahsaurabh0605, line 27: I am not sure how this is supposed to be used. I have raised an error when whenever there is a problem in `create`. Should I do this in all the files?

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,187 @@
...
+  def endpoint
+    @endpoint ||= begin
+      http = Net::HTTP.new(ENV["API_URL"] || 'rubybench.org', 443)
+      http.use_ssl = true
+      http
+    end
+  end
+
+  def generate_digest(path, database)
+    string = "#{File.read(path)}#{`ruby -v`}#{`rails -v`}"
```
tgxworld: Will this still work if we don't have `rails` installed now? Not that we only depend on ActiveRecord and Railties.

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,187 @@
...
+    request.set_form_data(form_data.merge(results))
+    endpoint.request(request)
+  end
+
+  def files
+    Dir["#{File.expand_path(File.dirname(__FILE__))}/*"].select! { |path| path =~ /bm_.+/ }
+  end
+
+  def run_single(path, connection: nil, database: nil)
+    script = "ruby #{path}"
```
tgxworld, line 107: Is there anything we need here to make ActiveRecord run in production mode?

sequel/benchmarks/support/benchmark_rails.rb
```diff
@@ -0,0 +1,54 @@
+require 'benchmark/ips'
```
tgxworld: Filename shouldn't be `benchmark_rails.rb
`

sequel/benchmarks/driver.rb
shahsaurabh0605, line 107: I checked with this and it is working. I don't think we need anything here.

sequel/benchmarks/driver.rb
shahsaurabh0605: Yes, I don't think it depends on rails anymore.

sequel/benchmarks/bm_sequel_destroy.rb
tgxworld, line 27: > I am not sure how this is supposed to be used

Try reading the documentation and read the source code if need be. A work around is not acceptable here.

sequel/benchmarks/driver.rb
tgxworld: Yea but we got to add Activerecord's version in here

sequel/benchmarks/bm_sequel_finders_find_by_attributes.rb
```diff
@@ -0,0 +1,34 @@
...
+DB.create_table(:users) do
+  primary_key :id
+  String :name, size: 255
+  String :email, size: 255
+  DateTime :created_at, null: true
+  DateTime :updated_at, null: true
+end
+
+class User < Sequel::Model
+  self.raise_on_save_failure = true
```
shahsaurabh0605, line 16: I think this should solve the problem.

sequel/benchmarks/driver.rb
```diff
@@ -0,0 +1,187 @@
...
+    @endpoint ||= begin
+      http = Net::HTTP.new(ENV["API_URL"] || 'rubybench.org', 443)
+      http.use_ssl = true
+      http
+    end
+  end
+
+  def generate_digest(path, database)
+    string = "#{File.read(path)}#{`ruby -v`}#{`grep activerecord Gemfile.lock | head -1`}"
+
```
shahsaurabh0605, line 141: Is a better way to get the version of our activerecord gem?

tgxworld, line 141: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/gem_version.rb

sequel/benchmarks/driver.rb
tgxworld, line 107: How did you check the ActiveRecord is running in production mode?

sequel/benchmarks/bm_sequel_finders_find_by_attributes.rb
tgxworld, line 16: :+1:

sequel/benchmarks/driver.rb
shahsaurabh0605, line 107: I meant I only checked only with `ruby #{path}` and not in production mode. If we cannot use `RAILS_ENV=production`, then how can we check for activerecord in production mode?

tgxworld, line 107: > then how can we check for activerecord in production mode?

That is what you need to find out :wink:

shahsaurabh0605, line 107: For activerecord to run in production mode in we need to first load the database YML file and establish our connection. Then we can use `APP_ENV=production` to run. How are we going to do this in our case?

tgxworld, line 107: > Then we can use APP_ENV=production to run.

Can you point me to documentation? I'll have a look

shahsaurabh0605, line 107: https://snippets.aktagon.com/snippets/257-how-to-use-activerecord-without-rails
Also, this answer on stack overflow: http://stackoverflow.com/a/37921998/5716507

tgxworld, line 107: Wait.. I just realized... Why do we even need ActiveRecord? We're not using it at all. Only ActiveModel is required.

You do need to take note that you can't just take stackoverflow answers as face value, please check your work before submitting

shahsaurabh0605:
I changed this to activemodel. I think it makes sense to run for rails in production environment. Now, when we have changed it to activemodel, is it any good to run in production environment?

sequel/benchmarks/bm_sequel_create_string_columns.rb
```diff
@@ -0,0 +1,32 @@
...
+DB = Sequel.connect(ENV.fetch('DATABASE_URL'))
+
+COUNT=25
+
+DB.create_table(:users) do
+  primary_key :id
+    COUNT.times do |i|
+      String :"column#{i}", size: 255
+    end
+    DateTime :created_at, null: true
```
tgxworld: please user 2 spaces for indentation

sequel/benchmarks/bm_sequel_scope_where.rb
```diff
@@ -0,0 +1,33 @@
...
+  DateTime :updated_at, null: true
+end
+
+DB.add_index :users, :email, unique: true
+
+class User < Sequel::Model
+  self.raise_on_save_failure = true
+end
+
+User.strict_param_setting = false
```
tgxworld: What is the rational behind setting this to false?

tgxworld:
I don't think ActiveModel has a production mode but do check the avaliable environment variables for Sequel to make sure it doesn't have one

sequel/benchmarks/bm_sequel_scope_where.rb
shahsaurabh0605: Running the sequel benchmarks without this gives [mass assignment error](http://sequel.jeremyevans.net/rdoc/files/doc/mass_assignment_rdoc.html).

tgxworld: Then we should define the setters

shahsaurabh0605: On using setters we have to set all the fields separately and also primary key separately. If there are more number of parameters this could be a problem. So why not use this?

shahsaurabh0605:
For setting the environment variable in sequel we need to set the production mode of our database.yml file as indicated [here](https://github.com/jeremyevans/sequel/blob/master/doc/bin_sequel.rdoc#yaml-connection-file). In our case I don't think it is of any use to run in production mode.

sequel/benchmarks/bm_sequel_scope_where.rb
tgxworld: Because in production, people will want to protect against mass assignment. Our benchmark here is to try and replicate as close as possible what others will do in production. :)

shahsaurabh0605: Done :+1:

sequel/benchmarks/bm_sequel_create_string_columns.rb
```diff
@@ -0,0 +1,31 @@
...
+  COUNT.times do |i|
+    String :"column#{i}", size: 255
+  end
+  DateTime :created_at, null: true
+  DateTime :updated_at, null: true
+end
+
+class User < Sequel::Model
+  self.raise_on_save_failure = true
+  self.set_allowed_columns *columns
```
tgxworld: Hmm where is columns being defined?

shahsaurabh0605: `columns` represents all the columns in the table.

sequel/benchmarks/bm_sequel_destroy.rb
```diff
@@ -0,0 +1,30 @@
...
+  primary_key :id
+  String :name, size: 255
+  String :email, size: 255
+  DateTime :created_at, null: true
+  DateTime :updated_at, null: true
+end
+
+class User < Sequel::Model
+  self.raise_on_save_failure = true
+  self.unrestrict_primary_key
```
tgxworld: Why are we setting this? We're not setting ids are we?

shahsaurabh0605: Yeah, this is for setting ids. Should I remove this?

tgxworld: Why did you add this?

shahsaurabh0605: As it is indicated [here](http://sequel.jeremyevans.net/rdoc/files/doc/mass_assignment_rdoc.html), to enable the use of primary key setters we need to add this.

tgxworld: Ok. 

> To enable use of primary key setters, you need to call unrestrict_primary_key for that model

Did you think about whether we need to set primary keys?

shahsaurabh0605: For our case id remains the primary key, so we may as well skip the setters for primary key.

tgxworld: Yup we are not setting the primary key at all so there is not need for that.

shahsaurabh0605: Done :+1:

tgxworld: cool thanks!

shahsaurabh0605:
Is there anything left to be done?

tgxworld:
Good work :) can you start looking at ruby-bench-docker so that we can get the benchmarks up on RubyBemch.org

shahsaurabh0605:
:+1:

PR #71, opened 2016-07-12
Added benchmarks to measure bundler's install, outdated, and lock per…

hmistry:
…formance.

Measures:
1. bundle install, using rails gem
2. bundle outdated, using rails gem
3. bundle lock, using rails gem (this tests lock after bundler has run once and cached the index)
4. bundle lock (initial), using rails gem (this is to test the performance of the initial run of bundle lock however it does not work as expected because the cache needs to be cleared between runs)

Also I've noted that the standard deviation reported is 0. After some investigation, I believe the measured times/iterations are probably ok, however the standard deviation value is fishy.

I've posted this PR for feedback.


bundler/benchmarks/bm_bundle_install_rails.rb
```diff
@@ -0,0 +1,19 @@
...
+
+Dir.mktmpdir do |dir|
+  Dir.chdir(dir) do
+    File.write('Gemfile', <<-GEMFILE)
+      source 'https://rubygems.org'
+      gem 'rails', '~> 5.0.0'
+      gem 'nokogiri', '1.6.7.2'
+    GEMFILE
+
+    system("bundle install --quiet")
```
segiddins, line 15: most of this time is going to be dominated by the network, unfortunately

hmistry, line 15: Noted. Would you know what the rough % breakdowns would be like 50% network, xx% ___ and xx% ___? Andre suggested using gemstash to reduce network uncertainty, which I'll try this week.

My thought for this one is it's a measure of what the user will experience. It's a sanity check for nothing was changed that causes a significant hit in overall performance without good reason and not for micro optimizations.

segiddins, line 15: The real issue is that network connection times and disk I/O dominate these operations, and they're going to vary a lot

indirect, line 15: You can completely remove the network from this by running `bundle pack` outside the benchmark, before running `bundle install`. In fact, we probably want one benchmark that installs a Gemfile of completely locked gems (for consistent comparison), and then a different benchmark that runs against a Gemfile without declared versions (to test the resolver, etc.).

At that point, I think the runtime will be dominated by the speed of unpacking the gems and copying the files—which is why it's good to have both `install` and `lock` in the benchmarks, so we can see changes/regressions in the resolver and the gem installation code.

hmistry, line 15: Ok, I'll give that a try and resubmit. Thanks.

hmistry:
@segiddins, @indirect I've updated the benchmarks with `install` having declared and undeclared versions in `Gemfile` (except for `nokogiri` due to installation issues). 

The intent for `bm_bundle_lock_gem_initial.rb` benchmark is to test `lock` on bundler's very first run when nothing is cached but I'm unable to reset the environment in between benchmark runs (tried using a custom suite but that only resets between warming and main run). Do you have any ideas for reseting the environment between each benchmark run to test the initial `lock` time? If not, I'll delete the file.

bundler/benchmarks/bm_bundle_install_rails.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    system("bundle package")
+    system("bundle install --quiet")
+
+    Benchmark.do_benchmark('bundle install rails', time: 60) do
+      system("bundle install --quiet")
+    end
+  end
+end
+
```
tgxworld: Small nitpick :scissors: the line please :)

hmistry:
@tgxworld Done.

hmistry:
Have you had a chance to review this PR? Any changes or ready to merge? I look forward to seeing the charts of Bundler's performance over time. 😄

bundler/benchmarks/bm_bundle_install_rails.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'support/benchmark_bundler.rb'
+require 'bundler/cli'
+
+# Fixed Nokogiri version because latest version 1.6.8 fails to install
```
tgxworld, line 4: Is 1.6.8 failing to install in the Docker image? If so, I'll have to fix that :smile:

hmistry, line 4: I haven't tested it with a Docker image. 1.6.8 fails on my local machine with all my Rails projects and there is an open issue on Nokogiri where many are facing problems with it, at last time I checked.

tgxworld, line 4: ah ok :)

tgxworld:
Thanks @hmistry!

hmistry:
@tgxworld When you (or someone on your team) have some time, could you look into why these benchmarks do not show on rubybench website?

tgxworld:
Hi @hmistry, I had a look but it seems like our bundler `driver.rb` is blowing up on one of our benchmarks causing the other benchmarks to not run :) It'll be great if you can have a look otherwise I'll fix it by this week. Thank you!

hmistry:
@tgxworld thanks for the update. I don't have access so I can't debug it. Let me know if there's any way I can help or how I can reproduce the setup on my local machine.

tgxworld:
@hmistry O you just have to run `driver.rb` :)

hmistry:
@tgxworld I looked into this a little and here's what I found.
It fails on:

```
ruby-bench-suite/bundler/benchmarks/support/benchmark_bundler.rb:60:
in 'do_benchmark': undefined method 'stddev_percentage' for #<Benchmark::IPS::Report::Entry:0x007fa0e403cd20> (NoMethodError)
```

This is due to the `benchmark-ips` regression which you reported https://github.com/evanphx/benchmark-ips/issues/76 and is supposedly fixed in v2.7.2. However, I'm running v2.7.2 on ruby 2.3.0 and it's still failing on my local machine!

If I comment the following line out, the benchmarks seem to run.
[benchmark_bundler.rb#L60](https://github.com/ruby-bench/ruby-bench-suite/blob/master/bundler/benchmarks/support/benchmark_bundler.rb#L60)

How did you resolve this with the other benchmark suites?

tgxworld:
Thanks for looking in to this. I locked `benchmark-ips` to `2.6.2` in our docker image

hmistry:
@tgxworld Yea, no problem. Anything I can do to help. So with that change, when will the benchmarks be available on the website? (It wasn't showing when I checked now)
Sorry to keep troubling you with this... I'm eager to see it. 😃

tgxworld:
O I meant that you'll need to lock benchmark-ips to 2.6.2 into order to get the suite to run. But one of the benchmark is still broken xD

hmistry:
Ok. Do you have an example of how I can lock benchmark-ips to 2.6.2 in the bundler suite? 
I'll look into the broken one. Thanks.

tgxworld:
@hmistry Hmm the only thing quick I can thing of is to uninstall all versions of `benchmark-ips` and run `gem install benchmark-ips -v 2.6.2`

tgxworld:
@hmistry Fixed in https://github.com/ruby-bench/ruby-bench-suite/commit/b79e41950c90b799c4e4b45f441197920c7344e6

PR #74, opened 2017-02-01
Add liquid benchmarks

jerryliu55:
The `bm_*` files call the benchmarking functions in https://github.com/Shopify/liquid/blob/master/performance/theme_runner.rb (`run` and `compile` for now, and `render` once https://github.com/Shopify/liquid/pull/851 is in).

Including liquid as a submodule to access its `performance/` directory which has the test liquid files and other useful things.

### DIFF
--- a/liquid_benchmarks/driver.rb
+++ b/liquid_benchmarks/driver.rb
@@ -1,0 +1,142 @@
+#
+# Liquid Benchmark driver
+#
+require 'digest'
+require 'net/http'
+require 'json'
+require 'pathname'
+require 'optparse'
+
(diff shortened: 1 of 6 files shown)


liquid_benchmarks/driver.rb
```diff
@@ -0,0 +1,142 @@
...
+
+      results.sort_by do |result|
+        result['iterations_per_second']
+      end.last
+    rescue JSON::ParserError
+      # Do nothing
+    end
+  end
+
+
```
tgxworld, line 117: Lets :scissors: the two lines above? :)

tgxworld:
@jerryliu55  Thank you!!!

PR #75, opened 2017-05-14
Sequel bm: Drop table if already exists when creating new

bmarkons:
After running sequal benchmark for the first time I have to manually drop table in order to run benchmark again.

In [activerecord benchmarks](https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/benchmarks/bm_activerecord_scope_all.rb#L10) table with the same name is being dropped before creating new with `force: true` option.

To achieve the same, I replaced [`create_table`](http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html#label-create_table) with [`create_table!`](http://sequel.jeremyevans.net/rdoc/files/doc/schema_modification_rdoc.html#label-create_table-21) in all sequal benchmarks.

### DIFF
--- a/sequel/benchmarks/bm_sequel_create_string_columns.rb
+++ b/sequel/benchmarks/bm_sequel_create_string_columns.rb
@@ -6,7 +6,7 @@ DB = Sequel.connect(ENV.fetch('DATABASE_URL'))
 
 COUNT=25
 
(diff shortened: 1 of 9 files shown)


tgxworld:
:+1:

PR #76, opened 2017-05-14
Add pg gem scope_all benchmark

bmarkons:
Add pg suite with `pg/postgres_scope_all` benchmark.

### DIFF
--- a/pg/benchmarks/pg_scope_all.rb
+++ b/pg/benchmarks/pg_scope_all.rb
@@ -1,0 +1,24 @@
+require 'bundler/setup'
+require 'pg'
+require_relative 'support/benchmark_pg'
+
+conn = PG.connect(
+  host: ENV.fetch("HOST", "localhost"),
+  port: ENV.fetch("PORT", "5432"),
+  dbname: ENV.fetch("DB_NAME", "rubybench"),
+  user: ENV.fetch("DB_USER", "postgres"),
+  password: ENV.fetch("DB_PASSWORD", "postgres")
+)
+
+conn.exec("DROP TABLE IF EXISTS users;")
+conn.exec("CREATE TABLE users ( name varchar(255), email varchar(255), created_at timestamp, updated_at timestamp );")
+
+1000.times do
+  conn.exec("INSERT INTO users VALUES ('Lorem ipsum dolor sit amet, consectetur adipiscing elit.', 'foobar@email.com');")
+end
+
+Benchmark.pg("raw_pg/postgres_scope_all", time: 5) do
+  str = ""
+  conn.exec("SELECT * FROM users").each_row{ |row| str << "name: #{row[0]} email: #{row[1]}\n" }
+end
+
--- a/pg/Gemfile
+++ b/pg/Gemfile
@@ -1,0 +1,15 @@
+source "https://rubygems.org"
+
+module ::Kernel
+  def pg_master?
+    ENV["PG_MASTER"] == '1'
+  end
+end
+
+if pg_master?
+  gem 'pg', github: 'ged/ruby-pg'
+else
+  gem 'pg', path: '/pg'
+end
+
+gem 'benchmark-ips', '~> 2.2.0'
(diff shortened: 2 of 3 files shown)


pg/benchmarks/pg_scope_all.rb
```diff
@@ -0,0 +1,23 @@
...
+conn.exec("DROP TABLE IF EXISTS users;")
+conn.exec("CREATE TABLE users ( name varchar(255), email varchar(255) );")
+
+1000.times do
+  conn.exec("INSERT INTO users VALUES ('Lorem ipsum dolor sit amet, consectetur adipiscing elit.', 'foobar@email.com');")
+end
+
+Benchmark.pg("raw_pg/postgres_scope_all", time: 5) do
+  str = ""
+  conn.exec("SELECT * FROM users").each_row{ |row| str << "name: #{row[0]} email: #{row[1]}\n" }
```
SamSaffron, line 22: interestingly this can be done a fraction quicker if you avoid the #each_row and instead just use a double while loop iteration cause you can avoid a bunch of block transitions, its very very minor

SamSaffron:
@tgxworld can we merge this?

pg/benchmarks/pg_scope_all.rb
```diff
@@ -0,0 +1,23 @@
+require 'bundler/setup'
+require 'pg'
+require_relative 'support/benchmark_pg'
+
+# Connect to POSTGRES database with provided values or use default
```
tgxworld: I'll prefer if we leave out the comment here. I believe the code is self explanatory.

pg/benchmarks/pg_scope_all.rb
```diff
@@ -0,0 +1,23 @@
...
+conn = PG.connect(host: ENV.fetch("HOST", "localhost"),
```
tgxworld: I prefer

```
conn = PG.connect(
  host: ENV.fetch("HOST", "localhost"),
  port: ..
)
```

The problem I have with flushing indentations is that changing anything in the first line forces us to re-indent all the lines polluting the commit history.

pg/benchmarks/pg_scope_all.rb
bmarkons: @tgxworld yeah, It is. I was kind of in doubt because of defaults. I'll leave it out :+1:

bmarkons:
@tgxworld, fixed stuff you've suggested :)

pg/benchmarks/pg_scope_all.rb
bmarkons: @SamSaffron interesing note :thinking:  I guess I should leave it as it is because I used iterating blocks in other scope_all benchmarks?

SamSaffron: yeah, it is a distraction, we should focus on getting that graph up on ruby bench, 3 lines on one benchmark

robin850:
As Noah mentioned in #79, there's a tiny schema inconsistency. Is that intentional that the `admin`, `created_at` and `updated_at` attributes aren't present in the PostgreSQL schema ? That would certainly be better to make sure that results can be compared apples to apples.

Looks great though ! 👍

bmarkons:
Nice catch! I omitted `created_at` and `updated_at` by mistake.

Although none of **scope_all** benchmarks has admin field in schema. Admin field is present only in **default_scope** benchmarks.

I am not sure if scope_all benchmarks are about to be compared with default_scope benchmarks, these two are of different type. I had only in mind comparing scope_all (or default_scope) across AR, sequel or pg implementations.

robin850:
Yeah, you're right Marko ! That looks great then !

bmarkons:
@tgxworld 
Do we want setup on RubyBench for pg and mysql gems same as for ActiveRecord and Sequel?
If yes, I think this one can be merged so I can continue with adding other benchmarks..

The next step would be to make and push docker image for pg suite.

tgxworld:
@bmarkons Sorry for missing this. I guess we should enable this since the benchmarks have been added :)

bmarkons:
Sure, I am writing driver for pg suite right now and Dockerfile.. :)

PR #77, opened 2017-05-15
Update sequel benchmark

jeremyevans:
Model.raise_on_save_failure has been true by default starting in Sequel 4.0.

Model.set_allowed_columns is deprecated and will be moved to a plugin in Sequel 5. There is no reason to set it in this case, since all attributes are allowed.

Model.all returns an array in Sequel, so the to_a is not necessary.

Note that Sequel does not automatically set timestamp values on the timestamp columns, and neither timestamp column has a default.  If you want something more similar to ActiveRecord behavior, you should use the timestamps plugin.  However, the recommended way to handle this when using Sequel is to use database triggers and default column values.

### DIFF
--- a/sequel/benchmarks/bm_sequel_scope_all.rb
+++ b/sequel/benchmarks/bm_sequel_scope_all.rb
@@ -13,8 +13,6 @@ DB.create_table!(:users) do
 end
 
 class User < Sequel::Model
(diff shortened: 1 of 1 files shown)


tgxworld:
:+1: Thank you @jeremyevans

PR #78, opened 2017-05-16
Make common benchmark runner to avoid duplication

bmarkons:
Benchmark groups `rails`, `sequel` and `liquid` have their own `support/benchmark_*.rb` for running, with exactly the same logic. I made common `benchmark_runner.rb` and placed it in `./support`.

 It is being called from those specific runners `*/support/benchmark_*.rb` with parameters required for certain group of benches.


noahgibbs:
+1, looks good.

Issue #12, opened 2015-03-25
Update bm_hash_aref_flo.rb


tgxworld:
https://github.com/ruby/ruby/commit/69a2df4eb6ad2e8066a9d0dd09c1bb869c49e9bd

Related: https://bugs.ruby-lang.org/issues/10999#change-51932

PR #79, opened 2017-05-16
Add sequel scope all with default scope bench

bmarkons:
Sequel default scope benchmark :chart_with_upwards_trend:

### DIFF
--- a/sequel/benchmarks/bm_sequel_scope_all_with_default_scope.rb
+++ b/sequel/benchmarks/bm_sequel_scope_all_with_default_scope.rb
@@ -1,0 +1,46 @@
+require 'bundler/setup'
+require 'sequel'
+
+require_relative 'support/benchmark_sequel'
+
+DB = Sequel.connect(ENV.fetch('DATABASE_URL'))
+
+DB.create_table!(:users) do
+  primary_key :id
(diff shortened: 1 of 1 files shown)


noahgibbs:
In general, this looks good. I don't think I see the "admin" boolean field in the schema for the pg and ActiveRecord flavors of this benchmark. Is that intentional?

bmarkons:
Actually admin field exists in schema for ActiveRecord **default scope** benchmark. Take a look at line [#12](https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/benchmarks/bm_activerecord_scope_all_with_default_scope.rb#L12). As opposed to **scope all** benchmarks of all flavors where admin field isn't added.

Do you suggest that scope all benchmarks and default scope benchmarks should have same schemas?

noahgibbs:
Ah, I probably read the wrong file, then. Sorry, if all the scope all benchmarks have the same schema then I was wrong. Carry on :-)

bmarkons:
Can I merge this one? This benchmark is comparable to https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/benchmarks/bm_activerecord_scope_all_with_default_scope.rb

PR #82, opened 2017-05-27
Add bm_erb_render.rb

k0kubun:
I want to add new benchmark for ERB, which is the same as https://github.com/ruby/ruby/blob/85fe3defcc71d40f09a6883b3db2fb5277b3df9c/benchmark/bm_erb_render.rb.

While I guess it's difficult, it would be helpful to backfill bm_erb_render's results for old commits if possible.

### DIFF
--- a/ruby/benchmarks/bm_erb_render.rb
+++ b/ruby/benchmarks/bm_erb_render.rb
@@ -1,0 +1,26 @@
+require 'erb'
+
+data = DATA.read
+max = 1_500_000
+title = "hello world!"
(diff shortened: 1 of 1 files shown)


bmarkons:
I think we could manually run only this one for number of previous commits. I could do it if @tgxworld think its fine.

tgxworld:
@k0kubun Thank you :+1:  

@bmarkons Maybe backfill this when you need to backfill other benchmarks? Otherwise, it'll take quite awhile just to compile all the rubies for a single benchmark.

PR #84, opened 2017-06-09
Don't reuse endpoints in sequel and rails drivers

bmarkons:
Fixing sequel and rails drivers same as https://github.com/ruby-bench/ruby-bench-suite/commit/53d5ef396429030f667c0d2e5f1f22702c30e247


tgxworld:
Thanks :+1:

PR #85, opened 2017-06-21
Fix selection benches to construct string from records

bmarkons:
As @SamSaffron suggested:

> This is kind of weak, it is not doing anything with the actual data, so if you defer materialize you automatically get a huge speed boost. We should probably always generate an actual string from these benchmarks to ensure that we materialize stuff, winning on deferred materialization is kind of winning by cheating. Yes, it happens sometimes in the real world but it is a bit of a crutch, why are you even selecting data you are not going to use?

This change will invalidate result data we have so far, but we need this fixed.

Take a closer look at discussion:
https://community.rubybench.org/t/how-can-active-record-be-faster/142

### DIFF
--- a/rails/benchmarks/bm_activerecord_finders_find_by_attributes.rb
+++ b/rails/benchmarks/bm_activerecord_finders_find_by_attributes.rb
@@ -27,5 +27,6 @@ end
 User.create!(name: 'kir', email: 'shatrov@me.com')
 
 Benchmark.rails("activerecord/#{db_adapter}_finders_find_by_attributes", time: 5) do
-  User.find_by_email('shatrov@me.com')
+  user = User.find_by_email('shatrov@me.com')
+  str = "name: #{user.name} email: #{user.email}"
 end
--- a/rails/benchmarks/bm_activerecord_scope_where.rb
+++ b/rails/benchmarks/bm_activerecord_scope_where.rb
@@ -25,6 +25,9 @@ class User < ActiveRecord::Base; end
 end
 
 Benchmark.rails("activerecord/#{db_adapter}_scope_where", time: 5) do
-  User.where(name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
-      .where("email LIKE :email", email: "foobar00%@email.com").to_a
+  str = ""
+  User
+    .where(name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
+    .where("email LIKE :email", email: "foobar00%@email.com").to_a
+    .each { |user| str << "name: #{user.name} email: #{user.email}\n" }
 end
--- a/sequel/benchmarks/bm_sequel_scope_all.rb
+++ b/sequel/benchmarks/bm_sequel_scope_all.rb
@@ -25,5 +25,8 @@ attributes = {
 end
 
 Benchmark.sequel("sequel/#{db_adapter}_scope_all", time: 5) do
-  User.all
+  str = ""
+  User.all.each do |user|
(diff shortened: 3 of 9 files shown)


rails/benchmarks/bm_activerecord_finders_find_by_attributes.rb
```diff
@@ -27,5 +27,6 @@ class User < ActiveRecord::Base; end
 User.create!(name: 'kir', email: 'shatrov@me.com')
 
 Benchmark.rails("activerecord/#{db_adapter}_finders_find_by_attributes", time: 5) do
-  User.find_by_email('shatrov@me.com')
+  user = User.find_by_email('shatrov@me.com')
+  str = "name: #{user.name} email: #{user.email}"
```
bmarkons, line 31: @SamSaffron do you think we need change here?

rails/benchmarks/bm_activerecord_scope_where.rb
```diff
@@ -25,6 +25,9 @@ class User < ActiveRecord::Base; end
 end
 
 Benchmark.rails("activerecord/#{db_adapter}_scope_where", time: 5) do
-  User.where(name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
-      .where("email LIKE :email", email: "foobar00%@email.com").to_a
+  str = ""
+  User
+    .where(name: "Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
+    .where("email LIKE :email", email: "foobar00%@email.com").to_a
+    .each { |user| str << "name: #{user.name} email: #{user.email}\n" }
```
noahgibbs, line 32: Could also do this with a map-and-join.

sequel/benchmarks/bm_sequel_scope_all.rb
```diff
@@ -25,5 +25,8 @@ class User < Sequel::Model
 end
 
 Benchmark.sequel("sequel/#{db_adapter}_scope_all", time: 5) do
-  User.all
+  str = ""
+  User.all.each do |user|
+    str << "name: #{user.name} email: #{user.email}\n"
+  end
```
noahgibbs, line 31: This, too, could be a map-and-join.

rails/benchmarks/bm_activerecord_scope_where.rb
SamSaffron: str << should be marginally faster, no need to carry an intermediate array

sequel/benchmarks/bm_sequel_scope_all.rb
SamSaffron: prefer the str << pattern here ...

rails/benchmarks/bm_activerecord_finders_find_by_attributes.rb
SamSaffron: yeah its legit imo.

rails/benchmarks/bm_activerecord_scope_where.rb
noahgibbs: Fair enough. I like map, but either way works fine.

rails/benchmarks/bm_activerecord_finders_find_by_attributes.rb
tgxworld: Do we need the `str` variable though?

SamSaffron:
I don't know... it probably makes no difference, you can look at what
instruction sequence mri creates, it may erase it ... not sure.

bmarkons:
Can I merge this one?

SamSaffron:
Sure

PR #86, opened 2017-06-22
Use sequel_pg

bmarkons:
We want to use `sequel_pg` for benchmarks with postgres. Since that is what people use mostly in production.

look at discussion:
https://community.rubybench.org/t/how-can-active-record-be-faster/142


tgxworld:
Looks good. Be sure to backfill the benchmarks :)

PR #87, opened 2017-06-27
Fix bench

bmarkons:


tgxworld:
@bmarkons Be sure to backfill this :)

PR #88, opened 2017-06-27
Add ActiveRecord benchmark as Discourse example

bmarkons:
This benchmark measure performance of real world query from Discourse.


bmarkons:
@SamSaffron Do you prefer having this query exactly the same as one from Discourse?

It would be much of a help if you could link to Discourse `schema.rb` file so I can use it to create exactly the same tables.

noahgibbs:
The Discourse query also isn't guaranteed to stay exactly the same. That's one problem with using real-world code in a benchmark - it changes over time.

SamSaffron:
This is sort of there... I would not worry about having it be 1-1 exactly the same as Discourse. This query is already fairly awesome in the way it makes AR work. 

I would just create the models in the bench, populate with some data and then generate a string that has about 30 topics worth of data in it. 

I would recommend reusing the "stage DB with data" script between AR and Sequel. 

Note that your sample is not using an includes on the users table yet, you are selecting a bunch of topics and you want to include the users (so AR side loads it) and you don't end up issuing an N+1 every time you grab a username.

bmarkons:
> Note that your sample is not using an includes on the users table yet, you are selecting a bunch of topics and you want to include the users (so AR side loads it) and you don't end up issuing an N+1 every time you grab a username.

Hmm, I am not sure what you're saying here. I was about to add tables needed to make this query work. Since users are being referenced from topics table I make users table. I need to add some fields though. 

Did you mean this - https://apidock.com/rails/ActiveRecord/QueryMethods/includes?

bmarkons:
@noahgibbs I moved staging DB with data in separate script which is about to be shared between Sequel and ActiveRecord benchmarks. I used ActiveRecord for staging data - is that fine? This setup is being run in separate process.

noahgibbs:
@bmarkons - yup, that sounds great. Thanks!

bmarkons:
I've added corresponding sequel benchmark.

I made sure same SQL query is being generated in both benchmarks.

BTW, I've run it locally and got these results:

Sequel:
- 136 ips
- 1623 objects

ActiveRecord:
- 97 ips
- 4001 objects

Do you find these benchmarks OK?

support/setup/bm_discourse_setup.rb
```diff
@@ -0,0 +1,148 @@
...
+
+# Stage DB
+
+10.times do |i|
+  User.create(username: "user#{i}")
+end
+
+User.all.each do |user|
+  500.times do |i|
+    user.topics.create(
```
tgxworld: We should user `create!` in this file. Otherwise, it might fail silently and we end up seeding incorrectly.

support/setup/bm_discourse_setup.rb
```diff
@@ -0,0 +1,148 @@
...
+    )
+  end
+end
+
+200.times do |i|
+  Category.create(topic: Topic.limit(1).offset(i).first)
+end
+
+Topic.last(100).each_with_index do |topic, i|
+  topic.category = Category.offset(i).limit(1).first
```
tgxworld: I think you need to save the record here or maybe just `update!`

bmarkons: oh yeah, nice catch :+1:

tgxworld:
Looks go to me :+1: Just one more question, I can't recall but do you know if we revert the fixtures after each benchmark?

bmarkons:
@tgxworld We are not doing anything after benchmark. We drop and create table before every benchmark in case it exists. But we don't touch other tables.

SamSaffron:
This is looking good, but I would like a few more things if possible.

- Can you confirm that all the right indexes are in place ... do an explain on the sql being generated, is it missing indexes? 

- Can you confirm that the text being generated by active record and sequel is exactly the same? 

- Can you make an extra benchmark for raw sql that generates the exact same text? 

The raw benchmark is very important cause it properly highlight ORM overhead.

bmarkons:
1. I haven't created any indexes in setup script. Here are both SQL queries generated :
https://pastebin.com/avVu5XYx

2. Yes, both text output are the same. Except retrieved datetime formats are slightly different:
`2017-07-05 13:22:23 +0000` | `2017-07-05 13:17:34 UTC`
Though I think we can't do anything about it? Here are both text outputs: https://pastebin.com/vUuT7Jng

3. Sure, I will work toward adding raw SQL benchmark.

SamSaffron:
We got to add indexes then, do an explain on the query.

Easiest here would be just to steal the indexes from Discourse.

We should aim at having format parity for dates, try to explicitly specify
format when you create the string.

bmarkons:
:sos: Little help needed.. How can I get indexes from Discourse?  I was looking for schema.rb file but no success.

noahgibbs:
I'm finding a db/structure.sql which appears to be a DB dump. Maybe use that?

bmarkons:
Thanks Noah. I did it :)

tgxworld:
@bmarkons We don't maintain a `schema.rb` in Discourse, you can look into the model files and there will usually be annotations at the bottom of each file which prints out the index.

bmarkons:
cool, I see it now. Thanks :clap:

bmarkons:
There's an obvious speed up after indexes were added.

### Query explanation :

Only difference between AR and Sequel queries is in column aliases. AR's `includes` generates column aliases like following: 
```
SELECT  
  "topics"."id" AS t0_r0,
  "topics"."created_at" AS t0_r1,
  "topics"."updated_at" AS t0_r2,
  "topics"."title" AS t0_r3,
  "topics"."bumped_at" AS t0_r4,
  "topics"."archetype" AS t0_r5
  ...
```
while Sequel don't use column aliasing there.
Do you think think we need any change here? That is how `includes` does internal bookkeeping and we want to compare that part to Sequel's `eager_graph` method therefore I think we should leave it as it is. 

Other than that everything in both generated SQL queries is identical :
- Left outer join with categories on category id.
- Left outer join with users on user id.
- Left outer join with topic_users on topic id and specific user id.
- Bypass deleted topics and private messages
- Only topics which are pinned globally
- Only topics with certain notification level set by user for category
- Order topics desc by bumped_at timestamp
- Limit to 30 topics

SamSaffron:
> Do you think think we need any change here? That is how includes does internal bookkeeping and we want to compare that part to Sequel's eager_graph method therefore I think we should leave it as it is. 

I think it is fine @jeremyevans  thoughts? 


Doesn't the user information though come from a second query in both cases?

jeremyevans:
Sequel only aliases columns when using eager_graph if the aliasing is actually needed (and tries to pick reasonable aliases in that case), so I think it makes sense to leave it as is.

bmarkons:
> Doesn't the user information though come from a second query in both cases?

```
...
AND (
  NOT EXISTS (
           SELECT 1 FROM category_users cu
             WHERE cu.user_id = 1
             AND cu.category_id = topics.category_id
             AND cu.notification_level = 0
             AND cu.category_id <> -1
             AND (tu.notification_level IS NULL OR tu.notification_level < 2)
   )
)
...
```

@SamSaffron This query is being generated as part of where condition.

SamSaffron:
The idea around include is that you defer the query. So:

```
[6] pry(main)> t = Topic.limit(10).includes(:user).to_a; puts
  Topic Load (0.7ms)  SELECT  "topics".* FROM "topics" WHERE ("topics"."deleted_at" IS NULL) LIMIT 10
  User Load (0.6ms)  SELECT "users".* FROM "users" WHERE "users"."id" IN (3840, 901, 27252, 1, 4424, 12705, 16828, 12311, 32, 27873)

[7] pry(main)> t.first.user.name
=> "Bob"
```

Notice how the .to_a issues the side load of users. 

I want to simulate this kind of behavior for both frameworks cause it is brutally expensive in Rails cc @jeremyevans

jeremyevans:
In Sequel::Model, eager does the second type of query.  The code for Sequel would be:

```ruby
t = Topic.limit(10).eager(:user).all
```

bmarkons:
In this bench AR uses JOINs for eager loading associations, therefore I've used `eager_graph` on Sequel side.

SamSaffron:
I follow but I am unsure if Discourse is doing: 

`.references(:user, :category)`

It does not make much sense there... its much smarter to load the users in a batch instead of making a giant row selecting users objects multiple times.

bmarkons:
I thought I couldn't reference users or categories tables in further query conditions if I don't tell AR to do join instead of preloading?

SamSaffron:
Yeah, but not following why are we using any columns in the users table here?

bmarkons:
Good point, we use only columns from categories :+1:

I'll make users be preloaded instead of joined on both AR and Sequel sides.

bmarkons:
Do you think this one can be merged? So I can switch to writing raw benchmark. I prefer doing it in separate PR.

SamSaffron:
absolutely go ahead and merge it, really really super extra mega keen to see some results on the rubybench

PR #89, opened 2017-06-29
Fix rails benchmarks

bmarkons:
This should fix mysql and scaffold benchmarks https://github.com/ruby-bench/ruby-bench-web/issues/193

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -28,7 +28,7 @@ end
 if mysql2_prepared_statements?
   # Does not work with < Rails 4.2.5.
   # See http://github.com/rails/rails/issues/21544.
-  gem 'mysql2', '0.4.2'
+  gem 'mysql2', '0.4.6'
 else
   gem 'mysql2', '0.3.18'
 end
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -33,6 +33,7 @@ else
   gem 'mysql2', '0.3.18'
 end
 
+gem 'tzinfo-data'
 gem 'pg', '0.18.1'
 gem 'benchmark-ips', '~> 2.2.0'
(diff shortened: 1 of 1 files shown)


noahgibbs:
Ah, nice! So the MySQL schema define error was just an outdated mysql2 gem?

bmarkons:
Exactly! Error is raised when activerecord tries to call `free` on `Mysql2::Result` but that method is introduced in version 0.4.3 :)

https://gemnasium.com/gems/mysql2/versions/0.4.3

tgxworld:
Awesome thanks for fixing @bmarkons

bmarkons:
@noahgibbs Do you think previous results should be invalidated when we change version of any dependency used? In this case mysql2.

https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/benchmarks/driver.rb#L145

^ here we generate digest and [compare it with previous](https://github.com/ruby-bench/ruby-bench-web/blob/master/app/models/benchmark_type.rb#L13) to check if benchmark file or ruby version has changed. Do you think we need to include [Gemfile](https://github.com/ruby-bench/ruby-bench-suite/blob/master/rails/Gemfile) in digest?

SamSaffron:
Yes, totally

noahgibbs:
I don't expect a significant performance difference between the two mysql2 gems. But I'm very willing to defer to Sam on this one. Invalidating the results *probably* doesn't matter, but if it matters, it matters a lot. So it's more conservative to re-run.

tgxworld:
Minor versions shouldn't have much impact so I would just observe the benchmark results under the new gem and re-run if necessary. Otherwise, I'm with @noahgibbs on this to just re-run it to be conservative.

PR #92, opened 2017-07-26
Add pg suite runner

bmarkons:
Current suite runner only runs all benchmarks without submitting results. I plan to do result submit in later PR. I want to have pg suite running in docker setup asap in order to debug [this](https://community.rubybench.org/t/make-comparison-chart-more-simple/149/9).

WDYT @tgxworld ?


tgxworld:
LGTM :+1:

PR #98, opened 2017-08-08
Reduce work in raw benchmark

bmarkons:
At first I've wanted to do sort of custom preloading of users in order to make fair comparison against ActiveRecord and Sequel.

Though I think it shouldn't be done in raw benchmark. WDYT?


SamSaffron:
Yeah the idea of raw is to use every trick in the books to properly measure
AR overhead

Issue #42, opened 2015-08-13
I think Discourse bench is broken


tgxworld:
results are not being posted.

SamSaffron:
ouch, we should try locally and see whats up

tgxworld:
@SamSaffron most likely due to the recent changes I made :stuck_out_tongue:

tgxworld:
```
rake aborted!
NoMethodError: undefined method `to_ary' for 86400:Fixnum
  (in /discourse/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb)
/discourse/lib/discourse_redis.rb:57:in `block (2 levels) in <class:DiscourseRedis>'
/discourse/lib/cache.rb:48:in `write_entry'
/discourse/app/models/emoji.rb:27:in `standard'
/discourse/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb:11:in `block in singleton class'
/discourse/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb:-5:in `instance_eval'
/discourse/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb:-5:in `singleton class'
/discourse/app/assets/javascripts/discourse/lib/emoji/emoji.js.erb:-7:in `__tilt_69944769901500'
Tasks: TOP => assets:precompile
(See full trace by running task with --trace)
Bundling assets
script/bench.rb:343:in `kill': no implicit conversion from nil to integer (TypeError)
    from script/bench.rb:343:in `ensure in <main>'
    from script/bench.rb:343:in `<main>'
```

Ruby commit `0970a928ef5079aa1e4745fe3a79af8685a82f46` onwards.

SamSaffron:
weird ... never seen that before, can you dump tmp directory and assets dir
and redis ?

tgxworld:
Fixed!

njt1982:
@tgxworld What did you do to fix it?

njt1982:
We fixed it. The problem we had was "launcher rebuilt [ID]" did a git pull and updated the launched script. This changed LOADS of stuff.

Most importantly:

a) It updated the base image which switched us from Ruby 2.0.0 to Ruby 2.3.0 (which caused this to_ary error **I THINK**).
b) As we were using the non-version locked postgres template, it caused our Postgres DB to be upgraded to 9.5).

tgxworld:
[...]

PR #102, opened 2017-08-16
Send pg suite results to web

bmarkons:
PG suite runner should send results to ruby-bench-web.

### DIFF
--- a/pg/benchmarks/pg_suite_runner.rb
+++ b/pg/benchmarks/pg_suite_runner.rb
@@ -21,7 +21,10 @@ class PGSuiteRunner
 
   def run
     bm_scripts.each do |bm_script|
-      run_benchmark(bm_script) if match_pattern?(bm_script)
+      if match_pattern?(bm_script)
+        output = run_benchmark(bm_script)
+        send_results(output, bm_script)
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


pg/benchmarks/pg_suite_runner.rb
```diff
@@ -40,8 +43,69 @@ def run_benchmark(bm_script)
...
+    send_ips(output, form_data)
+    send_objects(output, form_data)
+  end
+
+  def default_form_data(output, bm_script)
+    {
+      'benchmark_type[category]' => output['label'],
+      'benchmark_type[script_url]' => "#{RAW_URL}#{Pathname.new(bm_script).basename}",
+      'benchmark_type[digest]' => generate_digest(bm_script),
+      'benchmark_run[environment]' => "#{`ruby -v`}",
```
robin850: Nit-picky but you can certainly just do `ruby -v` ; it will directly return a string.

noahgibbs: And if you're sure you're running in the same Ruby, you can use the RUBY_VERSION constant, though that's not 100% equivalent to the above -- depends if you might be running in one Ruby while another is in the path :-)

bmarkons: Ah, thanks for noting this :+1: 

I think we need version of ruby from path since we are running benchmark scripts with it.
Here at line [#42](https://github.com/ruby-bench/ruby-bench-suite/pull/102/files#diff-bb13d860e242fc318e7fe26d9b203957R42)

I see now there's also `RUBY_DESCRIPTION` ~~variable~~ constant which returns same as `ruby -v` (patch level, revision etc) :smile:

PR #103, opened 2017-08-26
Fix bm_sequel_scope_all benchmark

bmarkons:
As @sgrif pointed out sequel benchmark was inserting nils for timestamps.

Sequel needs :timestamps plugin to support automatic update of timestamps.
@jeremyevans can you please verify this is the correct way to do it with Sequel? Thanks

### DIFF
--- a/sequel/benchmarks/bm_sequel_scope_all.rb
+++ b/sequel/benchmarks/bm_sequel_scope_all.rb
@@ -8,11 +8,12 @@ DB.create_table!(:users) do
   primary_key :id
   String :name, size: 255
   String :email, size: 255
-  DateTime :created_at, null: true
-  DateTime :updated_at, null: true
+  DateTime :created_at, null: false
+  DateTime :updated_at, null: false
 end
 
 class User < Sequel::Model
+  plugin :timestamps, update_on_create: true
 end
 
 attributes = {


sequel/benchmarks/bm_sequel_scope_all.rb
```diff
@@ -8,11 +8,12 @@
   primary_key :id
   String :name, size: 255
   String :email, size: 255
-  DateTime :created_at, null: true
-  DateTime :updated_at, null: true
+  DateTime :created_at, null: false 
+  DateTime :updated_at, null: false
```
robin850: Nit-picky but there's an extra space at the end of these lines.

jeremyevans:
The recommended way to implement created_at/updated_at timestamps is to use database defaults/triggers.  However, if you want something similar to what ActiveRecord does for comparison purposes, then the timestamps plugin does that.

bmarkons:
Yeah, I've wanted to make fair comparison with AR.

The context for benchmarks we have so far is narrowed to ruby side -- without database configuration etc. Though I understand it would be more efficient if you would set this up with triggers on db side. And we should probably add that benchmark, while leaving this one since most of Rails users are doing it this way.

noahgibbs:
Yeah. I don't know if most Sequel users use DB triggers for this. We didn't at Ooyala, even where we used Sequel. But nearly every ActiveRecord user does it Ruby-side.

PR #105, opened 2017-08-29
Add over_select benchmarks

bmarkons:
https://community.rubybench.org/t/can-sequel-be-configured-so-it-defer-materializes/159/9

### DIFF
--- a/pg/benchmarks/bm_scope_all_over_select.rb
+++ b/pg/benchmarks/bm_scope_all_over_select.rb
@@ -12,9 +12,11 @@ Benchmark.pg("pg/scope_all_over_select", time: 5) do
 
   result = connection.exec("SELECT * FROM users")
   result.type_map = type_mapper
-  
-  result.each_row do |row| 
-    str << "name: #{row[0]} email: #{row[1]} approved: #{row[2]} age: #{row[3]} birthday: #{row[4]} \n"
+
+  i = 0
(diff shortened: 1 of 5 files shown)


pg/benchmarks/bm_scope_all_over_select.rb
```diff
@@ -0,0 +1,20 @@
...
+
+Benchmark.pg("pg/scope_all_over_select", time: 5) do
+  str = ""
+
+  result = connection.exec("SELECT * FROM users")
+  result.type_map = type_mapper
+  
+  result.each_row do |row| 
+    str << "name: #{row[0]} email: #{row[1]} approved: #{row[2]} age: #{row[3]} birthday: #{row[4]} \n"
+  end
```
bmarkons: @SamSaffron will this do the work? Or you want to replace each blocks with while loop, even though we are using blocks in Sequel and Active Record benchmarks.

SamSaffron:
Raw should always be the fastest possible implementation, measure on local
see what is fastest

bmarkons:
Right, I'll switch to while loop then :racehorse:

bmarkons:
@SamSaffron pls verify this for me before I merge. Thanks

SamSaffron:
Looks good to me, let's get this backfilled

bmarkons:
Sure, I'm on it :+1:

PR #106, opened 2017-08-31
Remove unnecessary dependencies from Gemfiles

bmarkons:
I think we should keep our Gemfiles up to date for running latest Rails and Sequel versions.

Any additional setup for earlier versions we can customize in docker runner script.

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -26,15 +26,8 @@ else
   gem 'rails', path: '/rails'
 end
 
-if mysql2_prepared_statements?
-  # Does not work with < Rails 4.2.5.
-  # See http://github.com/rails/rails/issues/21544.
-  gem 'mysql2', '0.4.6'
-else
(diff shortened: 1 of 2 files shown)


tgxworld:
Hmm I'm not sure if this should be the responsibility of the docker runner script. Any reason why you'll like to move it there?

bmarkons:
I had in mind doing something like this for customizing Gemfile for lower versions of Rails: 
https://github.com/ruby-bench/ruby-bench-docker/blob/master/rails/master/runner#L39

What do you think?

tgxworld:
Ah we definitely want that our of the runner. My belief is that gem versions should be managed in the Gemfile. The runner script's sole responsibility is to just run the benchmarks.

bmarkons:
Fair enough, I am looking forward to move it out from runner :) closing this :no_entry:

PR #109, opened 2017-09-07
Support running rails 4.0.X versions

bmarkons:
@tgxworld This should serve as a workaround for conflicting `tzinfo` versions in rails 4.0.X.

https://community.rubybench.org/t/can-we-fill-up-our-benchmarks/153/13

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -8,6 +8,11 @@ module ::Kernel
   def mysql2_prepared_statements?
     ENV["MYSQL2_PREPARED_STATEMENTS"] == '1'
   end
+  
+  def v_4_0_x?
+    major, minor, micro = ENV["RAILS_VERSION"].to_s.split(".")
+    major == "4" && minor == "0"
+  end
 end
 
 if rails_master?
@@ -33,7 +38,10 @@ else
   gem 'mysql2', '0.3.18'
 end
 
-gem 'tzinfo-data'
+unless v_4_0_x?
+  gem 'tzinfo-data', '~> 1.2'
+end
+
 gem 'pg', '0.18.1'
 gem 'benchmark-ips', '~> 2.2.0'
 gem 'em-hiredis', '~> 0.3.0'


noahgibbs:
Using greater-than/less-than on strings for the version may have some odd quirks -- think of not-yet-existing versions like 4.0.10 or 4.10.2. I'm trying to find the documentation on the Rails object (e.g. Rails.env) - there may be a better way to get the preparsed version from that. I know there's a gem that will do this kind of thing, but that's a fairly large dependency for this small a check.

noahgibbs:
Hm. Nope, no obvious way to do it. What you want is something along the general lines of...

    def 4_0_x?
      major, minor, micro = Rails.version.split(".")
      major == "4" && minor == "0"
    end

That's a little clunky, but more accurate. You may see a way to express the same thing more elegantly :-)

bmarkons:
Thanks Noah. I'll make it as you proposed, it seems good to me.

rails/Gemfile
```diff
@@ -8,6 +8,11 @@ module ::Kernel
   def mysql2_prepared_statements?
     ENV["MYSQL2_PREPARED_STATEMENTS"] == '1'
   end
+  
+  def v_4_0_x?
+    major, minor, micro = ENV["RAILS_VERSION"].to_s.split(".")
```
bmarkons, line 13: It would be awesome if I could use `Rails.version` instead of `ENV["RAILS_VERSION"]` but it is not recognized here.

@noahgibbs Do you can think of any way to require Rails specified in Gemfile in order to use it in this function?

noahgibbs: Oh hey, that makes sense :-( Nope, that would be hard to do. You wouldn't want to depend on the Rails version being well known because this has to work even before the Rails gem is installed... You *could* make the Rails version be an in-Gemfile variable and use it both to specify the version of the Rails gem above and to check v_4_0_x?. But I don't know if that would be worth the trouble.

bmarkons: I'll leave it as it is right now then -- since we are passing `RAILS_VERSION` environment variable to [container](https://github.com/ruby-bench/ruby-bench-docker/blob/master/scripts/rails/releases.sh#L30) running suite.

noahgibbs: Sounds good to me.

PR #114, opened 2019-05-06
Fix discourse bench

OsamaSayegh:
Brings an up-to-date version of this script from discourse/discourse which fixes a couple problems.


tgxworld:
Thanks :+1:

PR #115, opened 2019-05-16
Fix RSS memory giving string instead of number

OsamaSayegh:


benchmark_driver-output-rubybench/lib/benchmark_driver/output/rubybench.rb
```diff
@@ -47,7 +47,9 @@ def create_benchmark_run(job, metric, context_result)
 
     result_hash = {}
     context_result.each do |context, result|
-      initiator_hash["benchmark_run[result][#{context.name}]"] = result.values.fetch(metric)
+      value = result.values.fetch(metric)
+      value = value.value if Struct === value && value.respond_to?(:value)
```
k0kubun, line 51: Not understanding what's going on here, but by plugin interface contract `BenchmarkDriver::Result#values` should be `{ BenchmarkDriver::Metric => Float (or Integer) }` https://github.com/benchmark-driver/benchmark-driver/blob/992961e1022dd25cde23f9283dbbe58a61bd053c/lib/benchmark_driver/metric.rb#L32.

Thus if `result.values.fetch(metric)` were Struct (not Float or Integer), that would be wrong and it should not be fixed in output plugin layer. For RSS memory, perhaps you should take a look at https://github.com/ruby-bench/ruby-bench-suite/blob/master/benchmark_driver-output-rubybench/lib/benchmark_driver/runner/rsskb.rb instead?

OsamaSayegh: @k0kubun thanks for the review. The problem here is that `result.values.fetch(metric)` was correctly returning a number ~2 weeks ago, but now it returns a Struct.

Looking at my PR again, I may have rushed a "fix" here instead of actually finding and fixing the root cause. I'll close this one and open another one in case I managed to find the root cause and fix it. Thanks again!

PR #116, opened 2019-05-17
Make RSS runner extract results correctly from `BenchmarkDriver::Repeater.with_repeat`

OsamaSayegh:
This commit https://github.com/benchmark-driver/benchmark-driver/commit/b235d963bc56ea272c740a1bf7b3f0d851403425 changed the return value of the `BenchmarkDriver::Repeater.with_repeat` method from a numeric value to a `Struct` instance, so we need to update the RSS runner to reflect the changes introduced by that commit.

cc @k0kubun

### DIFF
--- a/benchmark_driver-output-rubybench/lib/benchmark_driver/runner/rsskb.rb
+++ b/benchmark_driver-output-rubybench/lib/benchmark_driver/runner/rsskb.rb
@@ -44,11 +44,11 @@ class BenchmarkDriver::Runner::Rsskb
       jobs.each do |job|
         @output.with_job(name: job.name) do
           job.runnable_contexts(@contexts).each do |context|
-            value = BenchmarkDriver::Repeater.with_repeat(config: @config, larger_better: false) do
+            repeat_result = BenchmarkDriver::Repeater.with_repeat(config: @config, larger_better: false) do
               run_benchmark(job, context: context)
             end
             @output.with_context(name: context.name, executable: context.executable, gems: context.gems, prelude: context.prelude) do
-              @output.report(values: { METRIC => value }, loop_count: job.loop_count)
+              @output.report(values: { METRIC => repeat_result.value }, loop_count: job.loop_count)
             end
           end
         end
--- a/benchmark_driver-output-rubybench/lib/benchmark_driver/output/rubybench.rb
+++ b/benchmark_driver-output-rubybench/lib/benchmark_driver/output/rubybench.rb
@@ -33,8 +33,8 @@ class BenchmarkDriver::Output::Rubybench < BenchmarkDriver::BulkOutput
   # @param [BenchmarkDriver::Metric] metric
   # @param [Hash{ BenchmarkDriver::Context => BenchmarkDriver::Result }] context_result
   def create_benchmark_run(job, metric, context_result)
-    http = Net::HTTP.new(ENV.fetch('API_URL', 'rubybench.org'), 443)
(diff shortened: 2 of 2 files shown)


benchmark_driver-output-rubybench/lib/benchmark_driver/runner/rsskb.rb
```diff
@@ -44,11 +44,11 @@ def run(jobs)
       jobs.each do |job|
         @output.with_job(name: job.name) do
           job.runnable_contexts(@contexts).each do |context|
-            value = BenchmarkDriver::Repeater.with_repeat(config: @config, larger_better: false) do
+            repeat_result = BenchmarkDriver::Repeater.with_repeat(config: @config, larger_better: false) do
               run_benchmark(job, context: context)
             end
             @output.with_context(name: context.name, executable: context.executable, gems: context.gems, prelude: context.prelude) do
-              @output.report(values: { METRIC => value }, loop_count: job.loop_count)
+              @output.report(values: { METRIC => repeat_result.value }, loop_count: job.loop_count)
```
k0kubun, line 51: `run_benchmark` in this file seems to return Integer, and in that case `with_repeat` should be returning best or average `Integer`, not something having `#value`, at least in the latest benchmark_driver.gem. Could you elaborate more on this? Please explain the benchmark_driver version you're using too.

OsamaSayegh: Hi @k0kubun, thanks for the review

You're right, `run_benchmark` does return an integer, but it's called in a block that's passed to `BenchmarkDriver::Repeater.with_repeat` which does the following:

```ruby
      def with_repeat(config:, larger_better:, rest_on_average: :first, &block)
        values = config.repeat_count.times.map { block.call }

        desired_value =
          case config.repeat_result
          when 'best'
            best_result(values, larger_better)
          when 'worst'
            best_result(values, !larger_better)
          when 'average'
            average_result(values, rest_on_average)
          else
            raise "unexpected repeat_result #{config.repeat_result.inspect}"
          end
        RepeatResult.new(value: desired_value, all_values: extract_values(values))
      end
```
([method definition](https://github.com/benchmark-driver/benchmark-driver/blob/master/lib/benchmark_driver/repeater.rb#L19-L34))

Notice how it finds the desired value, and then wraps it inside an instance of a struct of type `RepeatResult` which is then returned. The return type of the `with_repeat` was changed to to struct ~2 months ago by the commit I linked above.

I'm using version `0.14.15` of benchmark_driver which is the latest version.

k0kubun: Oh... right. That's a really recent change and my local repository did not have that one when I wrote the above comment (sorry!). That makes a total sense.

k0kubun (approved):
Sorry for having this trouble by my change on benchmark_driver.gem :bow: And thank you for figuring out and fixing it!

OsamaSayegh:
No worries at all, thank you for merging!

At your convenience, could you please push a new version of this gem to rubygems.org? Thank you🙏

k0kubun:
Oh wow, I've forgotten how it's deployed and that it's a gem published to rubygems.org. I uploaded the current gem as v0.3.0 and also added @tgxworld as another owner of benchmark_driver-output-rubybench.gem.

OsamaSayegh:
Excellent, thank you so much!!

PR #118, opened 2019-05-21
`mem` is a hash of multiple memory metrics, not just RSS

OsamaSayegh:


tgxworld:
Hmm does `ruby-bench-web` know how to display metrics that are not `rss_kb`?

OsamaSayegh:
@tgxworld I think the JS chart library we use automatically builds a bar chart when there are multiple keys in the `results` hash.

image

tgxworld:
@OsamaSayegh O right I totally forgot about that. Thanks for this

